PR #48, opened 2018-10-03
[MAD DE] Francis Wardley

francisgeorge90:
Sorry I haven't finished. I will try and get it done when possible.


ta-web-mad:
No te preocupes Francis, es normal que cueste al principio. Sigue esforzándote al máximo 💪 

#Gabriel.

PR #55, opened 2018-10-18
PAR - Hanane done

HananeDUP:


starter-code/src/App.js
```diff
@@ -1,21 +1,234 @@
...
+import foods from './foods.json'
+
+class FoodTag extends Component {
+  constructor(props) {
+    super(props)
+    this.state = { qty: 1 }
+  }
+
+  handleChange(event) {
+    this.setState({ qty: parseInt(event.target.value) })
```
lucamerzi: It is a good practice to pass radix with parseInt like so:
```js
parseInt(string, radix)
```
Also, you have an alternative to parseInt(), which is Number()

starter-code/src/App.js
```diff
@@ -1,21 +1,234 @@
...
+    this.setState({ listFoods: arrayFood })
+  }
+
+  searchFilter(food) {
+    return food.name.includes(this.state.searchText)
+  }
+
+  addAppQuantity(name, val) {
+    // var arraySelected = {};
+    var selectedFoodCopy = this.state.selectedFood
```
lucamerzi: Technically, you are not making a copy here. In order to make a copy of the state you can use the slice method like so:
```js
```
But bravo for using the best practices of React (always making copy of state and never modifying it directly) 👏

ta-webpt-par:
Congrats on finishing the exo. 👍  This is one of the most challenging of the module! It sums up a lot of React features we've seen so far. Some points that came up during the exo:
```js

Number()


this.setState({ });


this.state = { counter: 1 }
    return {
    }
})


// setState IS ASYNCHRONOUS

    console.log(this.state)
})

```

HananeDUP:
Thanks Lucas for your comments! Interesting

Le jeu. 25 oct. 2018 à 00:03, ta-webpt-par <notifications@github.com> a
écrit :

PR #56, opened 2018-10-20
PAR - Robert Vivanco - iteration 2

rovisal:


ta-webpt-par:
Well done for the first iterations. Try to finish it up if you have some time to kill 🤓 . This is one of the most challenging exos of the module! It sums up a lot of React features we've seen so far. Some points that came up during the exo:
```js

Number()


this.setState({ });


this.state = { counter: 1 }
    return {
    }
})


// setState IS ASYNCHRONOUS

    console.log(this.state)
})

```

PR #57, opened 2018-10-20
PAR - Clément Sirvente

csirvente:
iteration 2

### DIFF
--- a/starter-code/src/App.js
+++ b/starter-code/src/App.js
@@ -1,8 +1,61 @@
-import React, { Component } from 'react';
-import logo from './logo.svg';
-import './App.css';
+import React, { Component } from "react";
+import logo from "./logo.svg";
+import "./App.css";
+import "bulma/css/bulma.css";
+import foods from "./foods.json";
+
+class Foodbox extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return (
+      <div className="box">
+        <article className="media">
+          <div className="media-left">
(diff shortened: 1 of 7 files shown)


starter-code/src/App.js
```diff
@@ -1,18 +1,46 @@
...
 class App extends Component {
+  constructor(props) {
+    super(props);
+    console.log(foods);
+    console.log(foods.slice(0, 5));
+    this.state = { listFood: foods.slice(0, foods.length) };
+  }
+
+  addNewFood (food) {
+    this.state.listFood.push(food)
```
lucamerzi: As a best practice, try to never modify the state object directly, by making copies, like so:
```js
```

ta-webpt-par:
Well done for the first iterations. This is one of the most challenging exos of the module! It sums up a lot of React features we've seen so far. Some points that came up during the exercise with you and your fellow students: 

```js
// Number()
You have an alternative to using parseInt() (which, in React.js is going to trigger some warnings)

this.setState({ });

this.state = { counter: 1 }


    return {
    }
})


// setState IS ASYNCHRONOUS

    console.log(this.state)
})

```

ta-webpt-par:
Well done for the first iterations. This is one of the most challenging exos of the module! It sums up a lot of React features we've seen so far. Some points that came up during the exo:

```js

Number()

this.setState({ });

this.state = { counter: 1 }


    return {
    }
})


// setState IS ASYNCHRONOUS

    console.log(this.state)
})

```

PR #58, opened 2018-10-20
PAR - Kevin Rabesaotra iteration 4 nearly done

knandraina:


starter-code/src/foodbox.js
```diff
@@ -0,0 +1,71 @@
...
+        super(props)
+
+        this.state = {
+          countItems : 1
+        }
+    }
+
+    AddCountItems = () => {
+      this.setState({
+        countItems : this.state.countItems + 1
```
lucamerzi: You should retrieve the value in the event object. You are always adding +1 to the state in charge of your input, which is why it always goes up one value once you click. Also, just so you know: 
```js
this.state = { counter: 1 }

	return {
	}
})
```

ta-webpt-par:
Congrats on getting pretty far into this exos. This is one of the most challenging exos of the module! It sums up a lot of React features we've seen so far. Some additional points that came up during the exo:
```js

Number()


this.setState({ });


this.state = { counter: 1 }
    return {
    }
})


// setState IS ASYNCHRONOUS

    console.log(this.state)
})

```

PR #98, opened 2018-12-05
[MAD DE] Eduardo Ruiz

ed1868:
Having A hard time with react. Need some review


ta-web-mad:
Well done Edu 🌴The exercise is not easy and it's almost finished and well implemented. Try to complete the iterations when you have some free time and if you have any doubts we can check it again.
Keep working like that 💪
#checked#closed#dʒiego 👌

PR #105, opened 2018-12-06
AMS Matt Hamers

mattviy:


IronhackAMS:
Nice Matt! 
Some remarks:
- because states of react elements in virtual DOM can change rapidly, React itself needs to follow up on these changes as well and see which elements are changed so that it can keep the DOM and virtual DOM in sync. To keep up with this you should use a unique key when using map methods as each array or iterator should have an unique key so try to implement keys to your lists.
- try to seperate your components in to seperated jsx files.
Kido

PR #110, opened 2018-12-07
AMS Dea

Deakristina:


IronhackAMS:
Well donre Dea! Small tip, when you are mapping through an array of objects and you have to deal with multiple elements you can also use the handy ES6 spread operator.

PR #123, opened 2018-12-12
AMS - Lyudmyla Baron

lxs07a:


IronhackAMS:
Looks good! Consider defining functions as methods of a class component, instead of defining them in the render method.

PR #130, opened 2018-12-12
MIA FTWD - Daniel Sauthier

DSauthier:


starter-code/src/Main.js
```diff
@@ -0,0 +1,146 @@
...
+  addTheFood = (e) => {
+    e.preventDefault();
+    if (this.state.nameInput === '' || this.state.caloriesInput === '' || this.state.imageInput === '') {
+      return;
+    }
+    const newFood = {};
+    newFood.name = this.state.nameInput;
+    newFood.calories = this.state.caloriesInput;
+    newFood.image = this.state.imageInput;
+    const copyOfFoodList = [...this.state.theList];
```
ta-web-mia: here, you have to make the copy of the FilteredFoods since that is the list that you use un the method showListOfFoods, like this:
```
const copyOfFoodList = [...this.state.filteredList];
    copyOfFoodList.unshift(newFood);

    this.setState({
      filteredList: copyOfFoodList,
      nameInput: '',
      caloriesInput: '',
      imageInput: ''
    });
```

starter-code/src/Main.js
```diff
@@ -0,0 +1,146 @@
...
+      const newFilteredList = fullList.filter((eachFood) => {
+        return eachFood.name.toLowerCase().includes(this.state.searchInput.toLowerCase())
+      })
+
+      this.setState({ filteredList: newFilteredList })
+    })
+  }
+
+
+  addToMenu = (foodToAdd, quantityToAdd) => {
```
ta-web-mia: in here, you have to first check if the food that you want to add is already in the todaysFoods list, if is not, you just need yo push, if it is you need to delete the one that was already there and push the new one (updated version)
```
    const copyOfList = [...this.state.todaysFoods];
    foodToAdd.quantity = quantityToAdd;
    let index = copyOfList.indexOf(foodToAdd)
    if( index > -1){
    copyOfList.splice(index,1)
    copyOfList.push(foodToAdd)
    this.setState({todaysFoods: copyOfList});
    } else {
    copyOfList.push(foodToAdd);
    this.setState({todaysFoods: copyOfList});
    }
```

ta-web-mia:
Good Job!!!! Only fix those two comments I added and it will be perfect!!!
💪🏼👨🏽‍💻

PR #148, opened 2019-02-20
Paris - Poeli & Christophe

grosbegue:
@pbojork


ta-web-paris:
Good use of the ternary condition! Great beginning of the lab!

PR #150, opened 2019-02-20
Paris: Alex, Anoop

brwnrcluse:
no css and search is case-sensitive

@alexjoly13


ta-web-paris:
That's great! Well done!

PR #160, opened 2019-02-21
Paris - Christina & Estelle

Olawinski-e:
Changed everything at the end because it was on the foodBox and had to be in the App.js
Not finished then.


Olawinski-e:
@cibailey

ta-web-paris:
I don't really know what to tell you as you did a lot but I can see that, as you said, you were lacking of time to rearrange everything to make it work. If you want, you can push it again later when you will have fixed what is not working. And you can tell us that you pushed it again! :)

PR #167, opened 2019-02-21
Paris-Bastien & Thomas

tomdreamer:
@basttran


starter-code/src/component/Search.js
```diff
@@ -0,0 +1,42 @@
...
+          foodFilter.toLowerCase()) !== -1
+      })
+      this.setState({
+        FilteredFood
+      })
+    }
+  
+    render() {
+      return (
+        <Poets FoodArray={this.state.FilteredFood} match={this.props.match} onChange={this.filterPoets} />
```
ta-web-paris: Where does this Poets come from? 🤷🏽‍♀️

starter-code/src/component/Search.js
```diff
@@ -0,0 +1,42 @@
...
+    constructor() {
+      super()
+      this.state = {
+        FoodArray: [],
+        FilteredFood: []
+      }
+    }
+  
+    componentWillMount() {
+      this.setState({
```
ta-web-paris: You should add : `const {FoodArray} = this.state;`

starter-code/src/component/Search.js
ta-web-paris: It can't work because there is no component called `Poets` and here you are calling it

ta-web-paris:
The main features are working. Only few bits are missing. Mainly that's a good work!

PR #170, opened 2019-02-21
Paris - Delphine & Francesca

wowfranzik:
done!

@Deie


starter-code/src/FormAddFood.js
```diff
@@ -0,0 +1,81 @@
...
+            Image:
+            <input
+              onChange={event => this.genericOnChange(event)}
+              value={this.state.image}
+              type="text"
+              name="image"
+              placeholder="image"
+            />
+          </label>
+          <button onClick={() => this.props.showForm()}>Submit</button>
```
ta-web-paris: If you put the function `showForm()` here, you won't have the function `handleSubmit()` working. So you should add the `showForm()` function inside the  `handleSubmit()` one.

ta-web-paris:
It's a good beginning. Only a small mistake was making the addition of new food impossible but you were very close! Good job!

PR #202, opened 2019-03-05
Mike and Monroe, iterarions 1 and 2 finished

MikeTroianello:


ta-web-mia:
Good job! There are a couple issues that should be addressed if you have the chance. The search bar differentiates between uppercase and lowercase, and there should be a quantity input for each food next to the button. However, the list is still appearing and the add food button works as well. Let me know if you need help with anything, and keep up the good work 👍

PR #204, opened 2019-03-05
Fischer - Jack - FTWD - Miami

fischerjack:
Meets all technical requirements. Could use a bit more CSS.


ta-web-mia:
Well done! Everything functions as intended, so I don't really have much to critique. Keep up the good work 👍

PR #221, opened 2019-03-30
complete iteration 1 and 2, and started iteration 3.

williamfiallos:


ta-web-mia:
Almost there! you've done everything right so far, now it's just a matter of making it dynamic to where you can add foods etc

PR #243, opened 2019-05-15
MIA FTWD ANGHY POMAR

apomarn:


ta-web-mia:
Nice job! The only problem I could find was that it was differentiating between capital and lowercase letters on the search function. For example, putting "a" in the searchbar would not make "Apple" show up. You could use a lowercase or uppercase method to make the search term and the name of the food more uniform, but it's a very minor issue. Overall, great job! Keep it up 👍

PR #244, opened 2019-05-15
MIA FTW April Parker Cooper

parkcoop:


ta-web-mia:
Nice job! The only issue I experienced is that the search function differentiates between uppercase and lowercase letters. It's a very minor issue though, so I wouldn't worry too much about it. Overall, fantastic job! Keep up the good work 👍

PR #245, opened 2019-05-15
MIA FTWD APR - Katy Brown

katybrown91:


ta-web-mia:
Nice job! Your search function works perfectly, and doesn't get messed up by capital letters or lowercase letters. I had some minor issues adding food to the list, as the quantity seems to be the calorie value. And quantities were overwriting each other rather than adding onto it. But overall, I think this app turned out well. Keep up the good work 👍

PR #246, opened 2019-05-15
MIA FTWD - Christian Khawam

CodeByCK:


ta-web-mia:
Nice job, all the functionalities work great! The only issue I ran into is that food quantities are overwriting each other instead of adding to it. For example, adding 1 pizza, and then clicking the button to add a 2nd pizza, will just overwrite the original and give you 1 pizza. Not a huge concern though, since it can just be seen as a way of finalizing the quantity. Nicely done, keep it up 👍

PR #249, opened 2019-05-16
MIA FTWD APR - Nathaniel Carvajal

nathancarv:


ta-web-mia:
Nice job! The search function works great, and the function to add new types of food works as well. Right now the only thing missing is the functionality to add the food to some sort of cart and to add the calories. Everything you have right now works perfectly though. Keep up the good work 👍

PR #250, opened 2019-05-16
MIA-FTWD-APR-Justin Velez

jvelez523:


ta-web-mia:
Nice job! The only thing that can really be improved on is having the food quantities add onto each other instead of replacing the current quantity. Nothing to get hung up over though. Very nice work, keep it up 👍

PR #251, opened 2019-05-16
MIA FTWD April - Chris Cardelle

ccardelle:


ta-web-mia:
Nice job! Your searchbar works very well, since it doesn't differentiate between uppercase and lowercase letters. I would suggest adding functionality to the quantity feature on both the food item and the list on the right, but everything else looks pretty good. Keep up the good work 👍

PR #252, opened 2019-05-16
IA FTWD APR 2019 Yulia

salapjaka:


ta-web-mia:
Well done! Aside from the bonus iterations, I would suggest that you use a form to display the images rather than file upload. The file upload doesn't seem to be working at the moment, but it's really no big deal since the name and calories forms work fine. Keep up the good work 👍

PR #257, opened 2019-05-21
MIA-FTWD-Allison

allisonkopp:


ta-web-mia:
Well done! One observation I had is that although the button to add the food is greyed out on the original food items, it isn't on the ones added by the user. Trying to add that food makes a NaN calorie and total appear in the cart, but it's quickly remedied by just adding a number amount of that food. Also, the remove food functionality seems to be removing the last item in the food array rather than it at a specific index. It's a bonus iteration though, so nothing to get too hung up on. Overall, great job! Keep it up 👍

PR #297, opened 2019-07-24
MIA-FTWD William Bennett

william1benn:


ta-web-mia:
Nice work on the search functionality, but make sure you sanitize the strings when you make a comparison. At the moment, capital letters are not equivalent to lowercase ones. Other than that, the only thing left to fix up is the shopping list/quantities. When you're setting state, make sure you console.log each step of the way to make sure values are being passed down where you expect them to. Keep up the good work, and take everything you've learned into your 3rd project. You've got this!
- Nick

PR #299, opened 2019-07-24
MIA FTWD - Andre Cabrera

andregcab:


ta-web-mia:
You're on the right track! I see that you're tracking e.target and e.value when you're setting the state. Just make sure you console.log the value to make sure information is getting passed through as intended. Keep up the good work, and let us know if you need any help.
- Nick

PR #337, opened 2019-10-01
MIA-FTWD-AUG2019-Elba Chimilio

elba3184:


ta-web-mia:
Obviously, going into the final project, it's not going to be necessary to keep working on this, but you should learn how to implement all of the functions that this completed app would have. There is a lot to be gained, in terms of understanding React.

PR #339, opened 2019-10-01
MIA-FTWD-AUG2019- Jimmy Thomas

hoshimotao:


ta-web-mia:
The search implementation works, but the add button doesn't and there seem to be some weird issues with adding food. At this point it's not worth it to keep working on the app, but try to understand how the mechanics of all of this works.

PR #344, opened 2019-10-02
[MIA-FTWD-102019] - Qiwei Lin

Kiwi-x-Kiwi:


ta-web-mia:
Most of this looks good, but the plus buttons don't actually add food to the total on the right, and when you add a new food, it just directly adds to the total, and doesn't create that as a future option to click.

PR #347, opened 2019-10-02
[MAD PP] JAVIER G FOMINAYA----- C. BORJA BERNARDEZ

borgila:
HARD SHIT! NO ES COÑA, PURO TIME TRAVELING

### DIFF
--- a/starter-code/package-lock.json
+++ b/starter-code/package-lock.json
@@ -316,9 +316,9 @@
       }
     },
     "async-each": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/async-each/-/async-each-1.0.1.tgz",
-      "integrity": "sha1-GdOGodntxufByF04iu28xW0zYC0="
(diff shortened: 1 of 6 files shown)


carlostrujillosamper:
Me he quedado loco ! Me ha transportado a mi niñez ..........

Aqui te dejo un extracto de "The Time Machine by HG Wells"

'Clearly,’ the Time Traveller proceeded, ‘any real body must have extension in four directions: it must have Length, Breadth, Thickness, and — Duration. But through a natural infirmity of the flesh, which I will explain to you in a moment, we incline to overlook this fact. There are really four dimensions, three which we call the three planes of Space, and  a fourth, Time. There is, however, a tendency to draw an unreal distinction between the former three dimensions and the latter, because it happens that our consciousness moves intermittently in one direction along the latter from the beginning to the end of our lives.’

‘That,’ said a very young man, making spasmodic efforts to relight his cigar over the lamp; ‘that... very clear indeed.’

PR #351, opened 2019-10-02
MIA Galindo Jose 2019

j-galindo:


ta-web-mia:
Most of it this is good, but you can't add the foods to a total, which adds all of the calories together. Otherwise, it looks good.

PR #352, opened 2019-10-02
MIA-FTWD-Enrique Narciso

e-narciso:
no bonus


ta-web-mia:
So a lot of this looks good, but there are some weird quirks with adding new items. They can't be searched for, and after a search is done, then removed, the item gets deleted. In addition, when adding the new food, it will add a different one instead. Otherwise, this looks very good.

PR #367, opened 2019-10-26
PTWD Berlin Chanty

ChantyHe90:


starter-code/src/App.js
```diff
@@ -13,6 +15,9 @@ class App extends Component {
         <p className="App-intro">
           To get started, edit <code>src/App.js</code> and save to reload.
         </p>
+        <ul>
```
hendrikniemann: This `<ul>` has no `<li>`. Even if it had some inside of `<FoodList>` the element should probably be inside of the food list component.

starter-code/src/components/AddFood.js
```diff
@@ -0,0 +1,79 @@
...
+    const inputName = event.target.name;
+    this.setState({
+      [inputName]: event.target.value
+    });
+  };
+  render() {
+    return (
+      <div>
+        <form onSubmit={this.handleFormSubmit}>
+          <label>name:</label>
```
hendrikniemann: Using the [for attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Attributes) of `<label>` will make your app more accessible.

starter-code/src/components/AddFood.js
```diff
@@ -0,0 +1,79 @@
...
+            value={this.state.image}
+            onChange={this.onFormFieldChange}
+          />
+          <label>quantity:</label>
+          <input
+            type="number"
+            name="quantity"
+            value={this.state.quantity}
+            onChange={this.onFormFieldChange}
+          />
```
hendrikniemann: We see this combination of `<label>` and `<input>` come up multiple times. This is a good opportunity to introduce a component.

PR #369, opened 2019-10-26
Berlin PTWD Christoph

cbut:


starter-code/src/App.js
```diff
@@ -1,19 +1,55 @@
...
+          <AddFoodForm onNewFood={this.addFood}></AddFoodForm>
+          : ""}
+        <SearchBar stateLocation={this.state}>
+
+        </SearchBar>
+        {this.state.foods.map((oneFood, idx) => {
+          return <div>
+            < FoodBox key={idx} {...oneFood} >
+            </FoodBox>
+          </div>
```
hendrikniemann: Why the extra `<div>` here? If you need it, I think it should be part of the `FoodBox` component. This might also error because the key should then be on the `<div>`.

starter-code/src/SearchBar.js
```diff
@@ -0,0 +1,32 @@
...
+            [inputName]: event.target.value
+        })
+        this.props.stateLocation(this.state)
+    }
+
+
+    render() {
+        return (
+            <div >
+                <form>
```
hendrikniemann: `<form>` is not needed here and can be omitted.

PR #371, opened 2019-10-27
PTWD-Berlin-Fairouz

fairouzw:


starter-code/src/App.js
```diff
@@ -1,21 +1,72 @@
...
+    });
+
+  searchFood = search => {
+    const results = [];
+    const foodList = [...this.state.foodData];
+    foodList.map(f => {
+      return f.name.toLowerCase().includes(search.toLowerCase())
+        ? results.push(f)
+        : " ";
+    });
```
hendrikniemann: You are abusing map as a loop here. I would either use a loop/foreach or use the filter method:

```js
const results = [];
this.state.foodData.forEach(f => {
  if (f.name.toLowerCase().includes(search.toLowerCase())) {
    result.push(f);
  }
});
// or
const results = this.state.foodData.filter(
  f => f.name.toLowerCase().includes(search.toLowerCase())
);

File: starter-code/src/App.js [hunk 2]
```diff
@@ -1,21 +1,72 @@
...
+          <h1 className="App-title">IronNutrition</h1>
         </header>
-        <p className="App-intro">
-          To get started, edit <code>src/App.js</code> and save to reload.
-        </p>
+        <Search searchFood={this.searchFood} />
+        <FoodList myfoods={this.state.foodData} addToToday={this.addToToday} />
+        <button className="button is-info" onClick={this.formVisibility}>
+          {" "}
+          Add new food{" "}
```
hendrikniemann: These `{" "}` white spaces here are coming from JSX not being able to handle whitespace before and after strings. But you don't need the whitespace there so I would delete them

starter-code/src/components/TodayFoods.js
```diff
@@ -0,0 +1,27 @@
...
+    return (
+      <div>
+        <div className="column is-half">
+          <div className="box">
+            <h1> Today's Foods </h1>
+            {this.props.foods.map((food, index) => (
+              <ul>
+                <li key={index}>
+                  {" "}
+                  {food.quantity} {food.name} = {food.calories} calories{" "}
```
hendrikniemann: Same with the `{" "}` here.

PR #372, opened 2019-10-29
PTWD - Berlin - Evelyn - iteration 2 done

Brezelfrau:


starter-code/src/App.js
```diff
@@ -1,18 +1,26 @@
...
+import logo from "./logo.svg";
+import "./App.css";
+import "bulma/css/bulma.css";
+import foods from "./foods.json";
+import FoodBox from "./FoodBox";
+import AddFood from "./AddFood";
 
 class App extends Component {
+  state = {
+    foods: foods
```
hendrikniemann: `foods` never changes. Therefore it would be better to not add it to the state.

PR #393, opened 2019-11-28
[PAR WEBDEV FT] Véro, Marion & Marie

Mawee33:
iterations 1 to 3 (with a lot of help)


ta-web-paris:
Good job for the first iterations ! It would be great to going on ;)

PR #396, opened 2019-11-29
[PAR - FTWD] Sarah

sarahledu:
done :)


ta-web-paris:
Good Job but be careful, the padding of your food list is over the form so it makes it hard to click on the inputs. Moreover, the form should disappear when you submit it.

PR #397, opened 2019-11-29
[PAR WDFT] Hanna

hanna974:


starter-code/src/App.js
```diff
@@ -1,18 +1,78 @@
...
+  addProduct = info => {
+    this.setState({ foods: [...this.state.foods, info], isActive: false });
+  };
+
+  handleChange = input => {
+    const searchValue = input.value;
+    let filtered = foods.filter(f =>
+      f.name.toLowerCase().includes(searchValue.toLowerCase())
+    );
+    this.setState({ foods: filtered });
```
ta-web-paris: for this lab you should not filter directly the state of foods otherwise, you loose the product you created before.

starter-code/src/App.js
```diff
@@ -1,18 +1,78 @@
...
+          <button className="addFoodBox" onClick={this.handleClick}>
+            Create FoodBox
+          </button>
+        </div>
+        <div className="form">
+          {this.state.isActive && <FormAddFoodBox clbk={this.addProduct} />}
+        </div>
+        <div className="columns">
+          <div className="column">
+            {this.state.foods.map((f, i) => (
```
ta-web-paris: you should implement the filter just before the map: this.state.food.filter(value of the search).map().
Thus you can keep all the food created.

And if the search is empty, it will display all the food.

starter-code/src/components/TodaysFoods.jsx
```diff
@@ -0,0 +1,12 @@
+import React, { Component } from "react";
+
+export default class TodaysFoods extends Component {
+  render() {
+    return (
+      <li>
+        {this.props.v.quantity} {this.props.v.name} ={this.props.v.calories}
```
ta-web-paris: it would be better to have the total calories displayed (cal * quantity)

ta-web-paris:
Nice Job, just some modification and it would be perfect ;)

PR #412, opened 2019-12-04
[MAD PP] Marvin, Rebecca

Rebeccawood:
we surrendered

### DIFF
--- a/starter-code/package-lock.json
+++ b/starter-code/package-lock.json
@@ -4,6 +4,21 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/runtime": {
+      "version": "7.7.4",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.7.4.tgz",
+      "integrity": "sha512-r24eVUUr0QqNZa+qrImUk8fn5SPhHq+IfYvIoIMg0do3GdK9sMdiLKP3GYVVaxpPKORgm8KRKaNTEhAjgIpLMw==",
+      "requires": {
+        "regenerator-runtime": "^0.13.2"
+      },
+      "dependencies": {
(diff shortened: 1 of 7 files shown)


ta-web-mad:
[:)](https://www.youtube.com/watch?v=EH3u9JYYZL4&start=91)

I think that you exhausted yourselves by taking on a way of doing the exercise that was more complex. With that said your code is actually really clean and well-structured.

I'm gonna try to explain how we can this form with the links.

This has changed and it's a bit different, so here is a solution:

1. We set a redirect that only renders when the form has submitted. When it renders, the page redirects. 
2. We control its rendering with a property of the state, this.state.submitted.

Form.js
```
class Form extends Component {
  constructor(props) {
    super(props);
    this.state = {
      name: "",
      calories: "",
      image: "",
      quantity: 0,
      submitted: false
    };
  }

  handleFormSubmit = e => {
    e.preventDefault();
    console.log(this.props);
    this.props.includesFood(this.state);
    this.setState({ submitted: true });
  };

[...]

render() {
    return this.state.submitted ? (
      <Redirect to="/" />
    ) : (
      <form onSubmit={this.handleFormSubmit}>
        <div className="field">
          <label className="label">Name</label>
          <div className="control">
          [...]
```

In app we modify the route in order make app the responsible of the data passed to the list:

With the props is how we pass the new list.
```
<Route exact path="/" render={() => <Foodbox newFoodList={this.state.food} food={this.state.food}></Foodbox>} />
```

In foodbox we paint the list using App's info insted of its own:

```
class Foodbox extends Component {
  constructor(props) {
    super(props);
    this.state = {
      food: this.props.food, // how do we bring newFoodList here?? // By grabbing the info from the props.
      BackUP: [...foods]
    };
  }
```

If any of you have any questions any questions feel free to come in the break.

Teo :turtle: :turtle:

PR #501, opened 2020-03-23
iterations 1-4 done

gha665:


gha665:
[PTWD102019] Evgeny Galitskiy

PR #574, opened 2020-05-03
done

fhoman:


starter-code/src/App.js
```diff
@@ -1,18 +1,123 @@
...
+         quantity: 0,         
+        }
+
+
+handleSearchInput = (e) => {
+const selectedFoods = this.state.foods.filter((food) => {
+      return food.name.toLowerCase().includes(e.target.value);
+    });
+this.setState({
+      selectedFoods:selectedFoods,
```
LloydC, line 35: Just wanted to point out that you can also write the short-hand syntax `this.setState({search:e.target.value, selectedFoods})`

starter-code/src/App.js
```diff
@@ -1,18 +1,123 @@
...
+
+handleInput = (e) => {
+  let {name, value} = e.target
+    this.setState({[name]: value})
+}
+
+addFood = () => {
+if (this.state.form === false) {
+this.setState({form:true})
+}
```
LloydC, line 48: I tried it with the following syntax and it works as well `addFood = () => {
  this.setState({form: !this.state.form})
}`

starter-code/src/App.js
```diff
@@ -1,18 +1,123 @@
...
+         <Search addFood={this.addFood} handleSearchInput={this.handleSearchInput} handleInput={this.handleInput} 
+         handleSubmit={this.handleSubmit} form={this.state.form} 
+         name={this.state.name} calories={this.state.calories} 
+        image={this.state.image} search={this.state.search}/>
+          
+           <div className='food-container'>
+         <div className='food-list'>
+     {this.state.selectedFoods.map((food,index)   =>
+     ( 
+      <FoodBox key={index} id={food.id} name={food.name} calories={food.calories} image={food.image}
```
LloydC, line 108: In a use case where you're passing all the properties of an object to a component, you might see/use the following syntax `<FoodBox {...food} key={index} handleInput={this.handleInput} addQuantity={this.addQuantity}/>`

starter-code/src/components/TotalsCheckout.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+export default function TotalsCheckout(props) {
+
+    console.log(props)
```
LloydC, line 5: Just wanted to mention that you can always check the values of your state and props with your React Dev Tools "Components". Nothing wrong with using console.log, just a friendly reminder ;)

PR #652, opened 2020-07-22
[BCN-PT-032020] Jesús Santana

jesussantana:
LAB | React IronNutrition

### DIFF
--- a/src/components/AddFood.js
+++ b/src/components/AddFood.js
@@ -1,0 +1,74 @@
+import React, { useState } from 'react';
+import { Form, FormGroup, FormControl, Button } from 'react-bootstrap';
+import { Link } from 'react-router-dom';
+
+const AddFood = (props) => {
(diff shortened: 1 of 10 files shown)


src/components/AddFood.js
```diff
@@ -0,0 +1,74 @@
+import React, { useState } from 'react';
+import { Form, FormGroup, FormControl, Button } from 'react-bootstrap';
+import { Link } from 'react-router-dom';
+
```
tomasyaya, line 4: const generateInitialState = id => ({
id: arrLength.toString(),
    name: '',
    calories: 0,
    image: '',
    quantity: 0,
})

useState(generateInitialState(arr.length.toString()))

const handleSubmit .....
`setLocalfood(generateInitialState((arrLength + 1).toString()))`

para no star escribiendo todo see objeto dos veces

PR #673, opened 2020-08-19
[WDFT-REMOTE-0820] Alejandro

alejandrohj:


src/App.js
```diff
@@ -1,26 +1,75 @@
...
+  }
+  handleDelete = (id) =>{
+    let toDelete = [...this.state.foodList];
+    toDelete.splice(id,1);
+    this.setState({
+      foodList: toDelete
+    })
+  }
+  render(){
+    return (<div style = {{display: 'flex',flexDirection: 'column', alignItems:'space-between'}}>
```
IH-WebDev-TA-Remote, line 49: for best practices, start your JSX files in a new line below the return, also watch for indentation as I see different ones used in this JSX. Use the commands for auto indentation to see the difference.

src/App.js
```diff
@@ -1,26 +1,75 @@
...
+    {!this.state.search && <button onClick = {() => {this.setState({search: true})}}>Add a new Food</button>}
```
IH-WebDev-TA-Remote, line 50: doing the setState here works fine but is recommended adding a handleClick function that does this. Helps with readability and reusability. For example in this case you could keep the button and click it again if you just want to hide the form again.

src/components/AddFood.js
```diff
@@ -0,0 +1,21 @@
...
+import { Form } from 'react-bootstrap';
+
+function AddFood(props){
+    let formStyle = {
+        display: 'flex',
+        flexDirection: 'column',
+        alignItems: 'center'
+    }
+    return (
+        <Form onSubmit = {props.onAdd} style = {formStyle}>
```
IH-WebDev-TA-Remote, line 12: although this works, try to get used to working with react forms that change the state dynamically and that is the information in the inputs value. It is good practices for thinking in react. After that you can call the function with the inputs that are in the state.

src/components/FoodBox.js
```diff
@@ -0,0 +1,50 @@
...
+            count: event.currentTarget.value
+        })
+    }
+    render(){
+        return (
+            <div className="box">
+                <article className="media">
+                    <div className="media-left">
+                    <figure className="image is-64x64">
+                        <img src={this.props.food.image} alt='img'/>
```
IH-WebDev-TA-Remote, line 21: when you have multiple props elements being passed, it is good practice to destructure them before the return to access the variables instead of doing props. ... everytime.

src/components/FoodBox.js
```diff
@@ -0,0 +1,50 @@
...
+                        <p>
+                        <strong>{this.props.food.name}</strong> <br />
+                        <small>{this.props.food.calories}</small>
+                        </p>
+                    </div>
+                    </div>
+                    <div className="media-right">
+                    <div className="field has-addons">
+                        <div className="control">
+                        <input onChange = {this.handleQuantity} className="input" type="number" value={this.state.count} />
```
IH-WebDev-TA-Remote, line 35: this is a perfect example of a dynamic react input that changes via the state, and the you call the onAddToday function with the state. So perfectly done :)

src/components/AddFood.js
alejandrohj: Hi Jorge!

I write you because I don't really know what's the meaning of 'react forms' in this feedback.

Sorry for the inconveniences,
Regards.

src/components/TodaysFood.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+function TodaysFood(props){
+    let todaysFood = props.foodList.filter((foods)=>{
+        return foods.quantity > 0
```
IH-WebDev-TA-Remote, line 5: this is actually a very clever way to go around the todays food iteration. Nice!

src/components/FoodBox.js
alejandrohj: Nicee! I'm gonna note it!

src/App.js
```diff
@@ -1,26 +1,75 @@
...
+
+  state = {
+    foodList: foods,
+    search: false
+  }
+  handleAddFood = (event) =>{
+    event.preventDefault();
+    let inputs = event.currentTarget
+    this.setState({
+      foodList: [{name: inputs.name.value, calories: inputs.calories.value, image: inputs.image.value, quantity:0}, ...this.state.foodList],
```
IH-WebDev-TA-Remote, line 21: suggestion for readability: create the object in a line above and then insert it in the state.

IH-WebDev-TA-Remote:
Very good work! all functionalities working. The final delete button was supposed to be for the today's list, not delete the foods from the main one, but still very good functionality to delete them both at the same time, very clever. 
Finally would be better if after clicking on the button to add you restarted the qty state to the default 1.
But again, very small things, overall excelent solution. Keep it up!

src/components/AddFood.js
IH-WebDev-TA-Remote: Similar to how you did in the FoodBox component for the quantity input. I will slack you one solution where you will see how these are implemented.

PR #674, opened 2020-08-19
[WDFT-REM-2007] Elise Dujardin

elisedjn:
All done! When you search for an ingredient and add one of the search result to the Today's Food, I've decided to re-render the all list of foods. I found that more practical regarding the UX, and I didn't know how to do it differently in the code XD


src/components/AddFood.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+const AddFood = (props) => {
+    return (
+        <form id="addFoodForm" onSubmit={props.onAdd}>
```
IH-WebDev-TA-Remote, line 5: calling the function and working around the target for the inputs works perfectly fine, but when working with forms in react is common to work around the state, and changing the inputs dynamically. Similar to how you did for the quantity in the FoodBox. then just invoke the function passing the information in the state.

src/components/FoodBox.js
```diff
@@ -0,0 +1,54 @@
+// Library
+import React from 'react';
+
+class FoodBox extends React.Component {
```
IH-WebDev-TA-Remote, line 4: indentation on some files set as 2 and in some set as 4. Remember to set all to 2. Not that important here but double check when working on your projects.

src/components/FoodList.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import Foodbox from './FoodBox'
+
+const FoodList = (props) => {
+    return (
+        <div className="column is-two-thirds">
+            {
+                props.foods.map((item, i) => {
+                    return <Foodbox
```
IH-WebDev-TA-Remote, line 9: if you are adding a JSX element in multiple lines it is better to start below the return, not in the same line. and  wrap it inside ( ).
return (
    <Foodbox 
    item={item} 
    key={i} 
    id={i} 
    onAdd={props.onAdd} 
    />
)

or like this:

        return <Foodbox item={item} key={i} id={i} onAdd={props.onAdd} />;

src/App.js
```diff
@@ -1,26 +1,103 @@
...
+      image: event.currentTarget.image.value,
+      calories: event.currentTarget.calories.value,
+      quantity: 0
+    }
+    this.setState({
+      foods: [newItem, ...this.state.foods],
+      filteredFoods: [newItem, ...this.state.foods],
+      addingFood: false
+    })
+    event.currentTarget.name.value = ""
```
IH-WebDev-TA-Remote, line 41: these 3 lines are not needed. The input fields from AddFood are being automatically reset because the component is being rendered again with no initial values set.

IH-WebDev-TA-Remote:
Very nice solution and even some cool styles, the bin icon in for the today's food was just lovely. Very good job! :) Keep it up!

PR #675, opened 2020-08-19
[WDFT-REM-2007] Britta Helm

brittahelm:
couldn't quite get the first bonus iteration one to work, submitting it now to go back to yesterday's bonus lab

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,104 @@
 import React from 'react';
-import logo from './logo.svg';
 import './App.css';
+import 'bulma/css/bulma.css';
+import foods from './foods.json';
+import FoodBox from './components/FoodBox';
+import AddFood from './components/AddFood';
+import Search from './components/Search'
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+class App extends React.Component {
+
+
+  state = {
+    foods: foods,
+    filteredFoods: foods,
+    todayFoods: [],
+    wantsToAdd: false
+  }
+
+  handleAddFood = (event) => {
+    event.preventDefault()
+    let foodName = event.currentTarget.food.value
+    let foodCal = event.currentTarget.calories.value
+    let foodImg = event.currentTarget.image.value
+    let newFoods = [{name: foodName, calories: foodCal, image: foodImg}, ...this.state.foods]
+    this.setState({
+      foods: newFoods,
+      wantsToAdd: !this.state.wantsToAdd
+    })
+  }
+
+  handleShowAdd = () => {
+    this.setState({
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -1,26 +1,104 @@
...
+  }
+
+  handleAddFood = (event) => {
+    event.preventDefault()
+    let foodName = event.currentTarget.food.value
+    let foodCal = event.currentTarget.calories.value
+    let foodImg = event.currentTarget.image.value
+    let newFoods = [{name: foodName, calories: foodCal, image: foodImg}, ...this.state.foods]
+    this.setState({
+      foods: newFoods,
```
IH-WebDev-TA-Remote, line 26: here you are updating the foods state but not the filteredFoods state. In L76 you are iterating over the filderedFoods, so the added items will only appear when you are also using the search feature.

src/App.js
```diff
@@ -1,26 +1,104 @@
...
+      filteredFoods: cloneFoods
+    })
+  }
+
+
+  handleAddToToday = (id, quantity) => {
+    let foodToAdd = this.state.foods[id]
+    foodToAdd.quantity = quantity;
+
+
```
IH-WebDev-TA-Remote, line 52: watch spacing

src/App.js
```diff
@@ -1,26 +1,104 @@
...
+
+
+
+
+  render(){
+    return (
+      <div className="App">
+      <div className="columns">
```
IH-WebDev-TA-Remote, line 69: watch indentation for JSX elements, these nested elements should be one step to the side. There is an auto-indentation feature in vs code that you could also use. But this kind of things help a lot for readability.

src/App.js
```diff
@@ -1,26 +1,104 @@
...
+      <div className="column">
+        {this.state.wantsToAdd && <AddFood onAdd={this.handleAddFood}/>}
+        {!this.state.wantsToAdd && <button onClick={this.handleShowAdd}
```
IH-WebDev-TA-Remote, line 72: if you want to show different things based on a boolean then a ternary is a better approach.

`this.state.wantsToAdd ? <AddFood onAdd={this.handleAddFood}/> : <button onClick={this.handleShowAdd} className="button">Add Food</button>`

src/components/AddFood.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import 'bulma/css/bulma.css';
+
+
+function AddFood(props) {
```
IH-WebDev-TA-Remote, line 5: this approach of passing the information of a form via the event is perfectly fine. But in react, while working with forms and inputs it is more common to change the values with the state. That way you can pass the information in the state to function as a parameter. See the solution sent to see what I mean. Or check how you did it for the quantity on FoodBox.

src/components/FoodBox.js
```diff
@@ -0,0 +1,38 @@
+import React, {useState} from 'react'
+
+
+
+export default function FoodBox(props) {
+
+    const [quantity, setQuantity] = useState(0);
```
IH-WebDev-TA-Remote, line 7: Amazing implementing hooks here! :)

IH-WebDev-TA-Remote:
Very good, most of the lab was done. There is an issue with the iteration for adding new foods, it is not updating the elements correctly. I pointed out the reason on the component. Other than that, all of the mandatory part is working perfectly. Keep it up!

PR #676, opened 2020-08-19
[Web-Rem-2007] Bodei

BodeiBrouwer:
Got stuck on iteration 5 and continued with yesterday's lab


src/components/Foods.js
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+import foodSeed from '../foods.json'
+import FoodBox from './FoodBox'
+import AddFood from './AddFood'
+import TodayFood from './TodayFood'
+
+
```
IH-WebDev-TA-Remote, line 7: remember to check spacing on your projects and set it to 2. Also, take a look at the auto-indentation feature of vs code with prettier to see how the code should be structured.

src/components/Foods.js
```diff
@@ -0,0 +1,122 @@
...
+    formAppear = () => {
+        this.setState({
+            formShow: true
+        })
+    }
+
+    handleAdd = (e) => {
+        e.preventDefault()
+        let foodName        = e.currentTarget.foodName.value
```
IH-WebDev-TA-Remote, line 27: interesting way to organize your variables. Looks good though :)

src/components/Foods.js
```diff
@@ -0,0 +1,122 @@
...
+        let foodCalories    = e.currentTarget.foodCalories.value
+        let foodImage       = e.currentTarget.foodImage.value
+        let foodQuantity    = e.currentTarget.foodQuantity.value
+        let newFood = [
+            {name:      foodName, 
+            calories:   foodCalories,
+            image:      foodImage,
+            quantity:   foodQuantity   
+            }, ...this.state.foodList]
+
```
IH-WebDev-TA-Remote, line 37: just for readability, this is the common way to write objects nested inside arrays.

```
let newFood = [
      {
        name: foodName,
        calories: foodCalories,
        image: foodImage,
        quantity: foodQuantity,
      },
      ...this.state.foodList,
    ];
```

src/components/Foods.js
```diff
@@ -0,0 +1,122 @@
...
+          foodList:       newFood,
+          filteredFoods:  newFood,
+          formShow:       false
+        })
+    }
+
+    handleSearch = (e) => { 
+      let searchName = e.currentTarget.value
+      let cloneFoods = this.state.foodList.filter((food) => {
+        return food.name.startsWith(searchName.toUpperCase())
```
IH-WebDev-TA-Remote, line 48: Currently it only works for one letter, after the second, it breaks. Try using the toUpperCase on the food.name as well. Also, it would be best to use .includes instead of .startsWith as the first one will check for all appearances. So you could try to find "roast" and get a result :)

src/components/Foods.js
```diff
@@ -0,0 +1,122 @@
...
+            foodList: foodSeed,
+            filteredFoods: foodSeed,
+            formShow: false,
+            TodayFoods: []
+        }
+    }
+
+    formAppear = () => {
+        this.setState({
+            formShow: true
```
IH-WebDev-TA-Remote, line 21: if instead of setting the formShow to be true you change it into the opposite of the current value, you could add the feature to toggle it on and off. Instead of having it appear only once.

src/components/Foods.js
```diff
@@ -0,0 +1,122 @@
...
+                </div>
+                <div className="column">
+                  <h3>Today's foods</h3>
+                  <ul>
+                  {
+                      this.state.TodayFoods.map((todayFood, i) => {
+                          return <TodayFood 
+                          todayFood={todayFood} 
+                          key={'todayfood' + i}
+                          id={i}
```
IH-WebDev-TA-Remote, line 110: if you are not using the id as a props you don't need to put it. key, on the other hand is something react uses internally, so that one should stay even if you don't use it later.

src/components/Foods.js
```diff
@@ -0,0 +1,122 @@
...
+        filteredFoods: cloneFoods
+    })
+    }
+
+    handleQuantity = (e) => {
+
+    }
+
+    handleToday = (e) =>{
+        let foodName        = e.currentTarget.food.name.value
```
IH-WebDev-TA-Remote, line 60: your target values are coming out as undefined, and that is the reason the code is breaking at this point. The reason is, for a button, the currentTarget will be the button (this is not a form button to submit info, it is a regular button). and that one doesn't hold the information of the food that should be added.
To go around this you could use state on the FodBox component to then pass the information of the food inside the function as a parameter :)

IH-WebDev-TA-Remote:
Good job, there was just a small thing breaking the search bar but everything else from the mandatory part was working correctly. Very good job! As for iteration 5 and on, there was a logic issue with the information being passed in the function. But overall good job! Check the solutions to see how those iterations are done and watch the logic of working with state on components that need to modify information or pass it to other components.

PR #677, opened 2020-08-19
[WEB-REM-2007] Lanette Lanting

llanting:
This lab was really hard, especially getting the quantity thing to work. In the end I managed to do it thanks to your tip of setting the state of quantity inside the component for foodbox. The bonus-iterations weren't too bad after that with some additional Google-ing. 

I tried adding an icon from Material Design Icons for Bonus 7, but found the documentation on Bulma not very helpful and I didn't get this to work.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,117 @@
-import React from 'react';
-import logo from './logo.svg';
+import React    from 'react';
+import foods    from './foods.json';
+import FoodList from './components/FoodList';
+import AddFood  from './components/AddFood';
+import Search   from './components/Search';
+import Calories from  './components/Calories';
+import 'bulma/css/bulma.css';
 import './App.css';
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+class App extends React.Component {
+
+  state = {
+    foods: foods,
+    filteredFoods: foods,
+    addShows: false,
+    todaysFoods: [],
+    total: 0,
+  }
+
+  handleAdd = (e) => {
+    e.preventDefault();
+    let name = e.currentTarget.name.value;
+    let calories = e.currentTarget.calories.value;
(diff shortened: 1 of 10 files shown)


src/App.js
```diff
@@ -1,26 +1,117 @@
...
+    this.setState({
+      foods: [{name: name, calories: calories, image, quantity: 0}, ...this.state.foods],
+      filteredFoods: [{name: name, calories: calories, image, quantity: 0}, ...this.state.foods],
+      addShows: false,
+    })
+  }
+
+  handleAddBtn = () => {
+    this.setState({
+      addShows: !this.state.addShows
```
IH-WebDev-TA-Remote, line 34: cool feature to toggle the form on and off!

src/components/AddFood.js
```diff
@@ -0,0 +1,34 @@
+import React    from 'react'
+
+function AddFood(props){
+    return (
+        <form onSubmit={props.onAdd}>
+            <div className="field">
```
IH-WebDev-TA-Remote, line 6: this way of passing information to the function via event.currentTarget works good. But in react it is more common working around the state. So having states for the different input fields that change dynamically with an onChange method. Then you can invoke the function with the information from the state.

src/components/Calories.js
```diff
@@ -0,0 +1,26 @@
+import React        from 'react'
+import CalorieItem  from './CalorieItem'
+
+function Calories(props){
+    return (
+        <div>
+            <h4 className="title is-4">Today's foods</h4>
+            <ul>
+                {
+                    props.todaysFoods.map((food, i) => {
```
IH-WebDev-TA-Remote, line 10: some indentation issues in this part (check default space set as 2). Also for returning multiline JSX it is better to start in a line below the return and wrap it in parentheses. 

```
return (
     <CalorieItem 
      key={i} 
      food={food} 
      id={i} 
      onDelete={props.onDelete} 
      />
);
```

src/components/FoodList.js
```diff
@@ -0,0 +1,22 @@
+import React    from 'react';
+import FoodBox  from './FoodBox'
+
+function FoodList(props){
+    return (
```
IH-WebDev-TA-Remote, line 5: same as Calories component, some small issues with indentation.

IH-WebDev-TA-Remote:
Great solution. Even when some nice styles and an even cooler delete button xD Very good job! 
Definitely the best approach for managing data on react is through state. it will help you render different information with changes and also to gather the info and send it to some other components or when invoking a function.
But again, overall very good solutions and a clean code. Keep up the amazing work :)

PR #678, opened 2020-08-19
[WEB-REM-2007] ALVARO

AlvaroSLam:
Lab done until the 5th iteration. I tried to have all the logic and do it similar to the second iteration, but it still give me some errors.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,127 @@
 import React from 'react';
-import logo from './logo.svg';
+import FoodBox from './components/FoodBox'
+import AddFood from './components/AddFood'
+import foods from './foods.json';
+import FoodList from './components/FoodList'
 import './App.css';
+import 'bulma/css/bulma.css';
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+
+class App extends React.Component {
+  constructor(props){
+    super(props)
+    this.state = {
+      foods: foods,
+      filteredFoods: foods,
+      formShow: false,
+      todayFoods: []
+    }
+  }
+  
+
+  handleAdd = (event) => {
+    event.preventDefault()
+    let foodName = event.currentTarget.foodName.value
+    let foodCalories  = event.currentTarget.foodCalories.value
+    let foodImage = event.currentTarget.foodImage.value
+
+    let newFood = [
+      {
+        name: foodName,
+        calories: foodCalories,
+        image: foodImage,
+      }, ...this.state.foods]
+
+    this.setState({
+      foods: newFood,
+      formShow: false
+    })
+  }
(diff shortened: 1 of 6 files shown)


src/components/AddFood.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+function AdFood(props){
+    return(
```
IH-WebDev-TA-Remote, line 4: a simple submit and sending the input values via the event target is ok, however, using in React you could use the state for getting the input values and sending them on the function invocation via parameter. Check the solution sent to see how it is done there.

src/App.js
```diff
@@ -1,26 +1,127 @@
...
+
+  handleToday = (event) =>{
+    let foodName = event.currentTarget.food.name.value
+    let foodCalories = event.currentTarget.food.calories.value
+    let foodQuantity = event.currentTarget.food.quantity.value
+
+
+    let newFood = [
+        {
+
```
IH-WebDev-TA-Remote, line 68: check for unneeded spacing.

src/App.js
```diff
@@ -1,26 +1,127 @@
...
+  }
+  
+
+  handleAdd = (event) => {
+    event.preventDefault()
+    let foodName = event.currentTarget.foodName.value
+    let foodCalories  = event.currentTarget.foodCalories.value
+    let foodImage = event.currentTarget.foodImage.value
+
+    let newFood = [
```
IH-WebDev-TA-Remote, line 28: the variable name newFood implies this is one food element, but you are getting all the foods array (the new + all the other). Be mind with naming your variables as it really helps with readability.

src/App.js
```diff
@@ -1,26 +1,127 @@
...
+  }
+
+  handleSearch = (event) => {
+    let searchName = event.currentTarget.value
+    let cloneFoods = this.state.foods.filter((item) => {
+      return item.name.startsWith(searchName)
+    })
+
+    this.setState({
+      filteredFoods: cloneFoods
```
IH-WebDev-TA-Remote, line 54: you are updating the filteredFoods state here but in line 97 you are mapping over the foods state. That is why the filter functionality is not working.

src/App.js
```diff
@@ -1,26 +1,127 @@
...
+  handleShowForm = () => {
+    this.setState({
+      formShow: true
+    })
+  }
+
+  handleSearch = (event) => {
+    let searchName = event.currentTarget.value
+    let cloneFoods = this.state.foods.filter((item) => {
+      return item.name.startsWith(searchName)
```
IH-WebDev-TA-Remote, line 50: instead of startsWith you can use .includes. That one will do a better search functionality. Also convert both strings to either upperCase or lowerCase so it is not case sensitive. Check solution sent.

src/App.js
```diff
@@ -1,26 +1,127 @@
...
+            placeholder="Search">
+          </input>
+
+         
+        
+        {
+          this.state.foods.map((food, index) =>{
+            return(
+            <FoodBox  
+            food={food}
```
IH-WebDev-TA-Remote, line 100: when mapping to render components, always add a key property to the component that should be unique for each element. Usually this uses the the index being passed in the map.

IH-WebDev-TA-Remote:
Good job, even though the search feature was not working it was only due to a simple mistake. check the notes in the method. Also seems you tried to do the bonus. Go through the solution to see how to do that and if you have the time, try to go over those again. Keep it up!

PR #679, opened 2020-08-19
[WEB-REM-2007] Rafael

rlaz133:
I think this solution for iteration 5 is not the most elegant, but I was dense today and stuck to things I knew would work. Also, sending back props to the parent component in the arguments of the function triggered a depth error.

For the search bar, it will first give the results that begin with the string typed and then the results that contain the string typed, as a sort of fuzzy match.


src/App.js
```diff
@@ -1,26 +1,70 @@
...
+      filteredFoods: foods,
+      todayFoods:[],
+      displayForm: false
+    }
+  }
+
+  toggleForm=()=>{
+    let formState = this.state.displayForm
+    formState = !formState
+    this.setState({displayForm: formState})
```
IH-WebDev-TA-Remote, line 25: you could reduce code here, just doing this single line

`this.setState({displayForm: !this.state.displayForm})`

src/components/AddForm.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+export default function AddForm (props){
+  return(
+    <form onSubmit={props.addFood}>
```
IH-WebDev-TA-Remote, line 5: working with a simple form and passing the information through the event is fine. But in react you can use the state to dinamically change the values of the input field. And then use the information in the state to use it on the function call via parameter.

src/components/FoodBox.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  
+  render(){
+  return(
+    <div className="box">
+      <article className="media">
+        <div className="media-left">
+          <figure className="image is-64x64">
+            <img src={this.props.food.image} alt='' className='food-pic'/>
```
IH-WebDev-TA-Remote, line 31: if you have multiple variables passed from props or the state. It is good practice to destructure them before the return and then use the individual variables.

src/App.js
```diff
@@ -1,26 +1,70 @@
...
+  }
+
+  addFood=(event)=>{
+    event.preventDefault();
+    let newFood = {name: event.currentTarget.name.value, calories: event.currentTarget.calories.value, image: event.currentTarget.image.value}
+    this.setState({foods: [...this.state.foods, newFood], filteredFoods: [...this.state.foods, newFood]})
+  }
+
+  searchFood=(event)=>{
+    let greatMatches = this.state.foods.filter((food)=>{return food.name.toLowerCase().startsWith(event.currentTarget.value.toLowerCase())})
```
IH-WebDev-TA-Remote, line 35: I believe doing only .includes also includes instances where the strings starts with the characters in the input, so doing both filters might be a little redundant. Just using the .includes should work the same.

IH-WebDev-TA-Remote:
Very nice solution and cool styles. One of the bonus iterations was to add to the existing one if it was already in the list. As not to have any repeated food items. But other than that everything is working great and the code is very reduced and very easy to read. Amazing work :)

PR #680, opened 2020-08-19
lab nutrition

GunnerAg:
was not able to finish the 5th iteration, I wanted to use a new component to render the table with the values but I get the feeling Im making it way more complex than it really is !

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,91 @@
 import React from 'react';
-import logo from './logo.svg';
+import foods from './foods.json';
+import FoodBox from './components/FoodBox';
+import MenuDay from './components/MenuDay';
+import AddFood from './components/AddFood';
+import Searchbar from './components/Searchbar';
 import './App.css';
+import 'bulma/css/bulma.css';
+import ButtonAddForm from './components/ButtonAddForm';
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+
+class App extends React.Component{
+
+  constructor(props){
+    super(props)
+    this.state ={
+        foods: foods,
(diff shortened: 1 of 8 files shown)


src/components/AddFood.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+function AddFood(props) {
+
+    return (
+        <form onSubmit={props.onAdd}>
```
IH-WebDev-TA-Remote, line 6: using regular forms and passing the information via the event works fine. However in react, you can also work around the state, having the input values as state elements and updating the input tags dynamically onChange. Then use the state values to pass the information when you call the function.

src/App.js
```diff
@@ -1,26 +1,91 @@
...
+   let newFoods =[{name:newFood,calories:newCalories,image:newImage},...this.state.foods]
+   this.setState({
+     foods:newFoods,
+     filteredFood:newFoods
+
+   })
+
+ };
+ hideOrShow=()=>{
+   this.setState({condition:!this.state.condition})
```
IH-WebDev-TA-Remote, line 39: cool feature to toogle the forms based on clicks :)

src/App.js
```diff
@@ -1,26 +1,91 @@
...
+   let clonedFoods = this.state.foods.filter((food)=>{
+     return food.name.toLowerCase().includes(foodToSearch.toLowerCase())
+   })
+   this.setState({
+     filteredFood:clonedFoods
+   })
+
+ };
+
+ handleAdding=(a, b, c)=>{
```
IH-WebDev-TA-Remote, line 54: this is good approach for iteration 5. however, instead of adding the elements in an array. Best apprach would be to create one object with all properties (name, amount, calories) and then insert the object into the todaysFood state. That way when you map over the todayObject array in L67, you know each element of the map will be each food object added to the array.

src/components/FoodBox.js
```diff
@@ -0,0 +1,106 @@
...
+            amount:newCounter
+        })
+    }
+
+
+
+    render() {
+        return (
+        
+
```
IH-WebDev-TA-Remote, line 27: check for unneeded spacing

IH-WebDev-TA-Remote:
Good solution up until iteration 5. And even there you had most of the code to show the information. Check the notes on the method to see how the rest should have been handled, or even check the solution sent. But again, very good job, keep it up!

PR #681, opened 2020-08-20
[WEB-REM-2007] Marta Lourido

MartaLourido:
I couldn't finish it, I did until interaction 3 and I hope to finish it today

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -3,6 +3,8 @@ import './App.css';
 import 'bulma/css/bulma.css';
 import FoodBox from './components/FoodBox'
 import foods from './foods.json';
+import AddFood from './components/AddFood'
+import FoodForm from './components/FoodForm'
 
 // import DisplayFood from './components/DisplayFood'
 
@@ -26,8 +28,13 @@ class App extends React.Component {
     })   
   }
 
-  handleQuantityChange = () => {
-
+  handleQuantityChange = (event) => {
+    const { name, value, quantity } = event.target;
+    this.setState({
+      name: name,
+      value: value,
+      quantity: quantity
+    })   
   }
 
     render(){
@@ -35,9 +42,14 @@ class App extends React.Component {
       <div className= 'IronContacts'>
       <h2>IronNutrition</h2>
       <button onClick={() => this.handleAddfood()}>Add new food</button> 
-      <FoodBox theFood={foods} onQuantityChange={this.handleQuantityChange} />
+      {foods.map((item, index) => {
+        return (
+      <FoodBox key={index} theFood={item} onQuantityChange={this.handleQuantityChange} /> 
+      )})}
+     
       </div>
     );
   }
 }
+
 export default App;
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,43 @@
 import React from 'react';
-import logo from './logo.svg';
 import './App.css';
+import 'bulma/css/bulma.css';
+import FoodBox from './components/FoodBox'
+import foods from './foods.json';
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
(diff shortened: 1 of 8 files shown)


src/App.js
```diff
@@ -1,26 +1,80 @@
 import React from 'react';
-import logo from './logo.svg';
 import './App.css';
+import 'bulma/css/bulma.css';
+import FoodBox from './components/FoodBox'
+import foods from './foods.json';
+import AddFood from './components/AddFood'
+import TodaysFoods from './components/TodaysFoods'
+import FoodForm from './components/FoodForm'
+import searchBar from './components/searchBar'
```
IH-WebDev-TA-Remote, line 9: components needs to be written in PascalCasing. SearchBar. This is very important, otherwise the JSX won't recognise them.

src/App.js
```diff
@@ -1,26 +1,80 @@
...
+  this.setState({
+    showForm: true,
+  })
+}
+
+    render(){
+    return (
+      <div>
+      <div className= 'IronContacts'>
+      <h2>IronNutrition</h2><searchBar filteredFood={this.filteredFood} />
```
IH-WebDev-TA-Remote, line 62: same as above. SearchBar needs to be in PascalCasing

src/App.js
```diff
@@ -1,26 +1,80 @@
...
+      
+      <FoodForm handleChange={this.filteredFood} name='search' type='text' />
```
IH-WebDev-TA-Remote, line 64: you have two search bars. The one above that was not working because of how it was written, and this one that is not doing anything.

src/components/FoodForm.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const FoodForm = (props) => {
```
IH-WebDev-TA-Remote, line 3: unneeded component, as you already have the SearchBar one

src/components/searchBar.js
```diff
@@ -0,0 +1,31 @@
+import React, { Component } from 'react';
+
+class searchBar extends Component {
+  state = {
+    input: '',
+  };
+  handleChange = (event) => {
+    const { value } = event.target;  
+    this.props.filteredFood(event.target.value);
```
IH-WebDev-TA-Remote, line 9: the component is working correctly. However, here you are passing event.target.value. In line 8 you deconstructed the value from event.target, so in this line you can only put value and it will work.

src/App.js
```diff
@@ -1,26 +1,80 @@
...
+    const { name, value, quantity } = event.target;
+    this.setState({
+      name: name,
+      value: value,
+      quantity: quantity
+    })   
+  }
+
+  filteredFood = (e) => {
+    const searchText = e.currentTarget.value
```
IH-WebDev-TA-Remote, line 43: when you call this function in the SearchBar component you are passing the string value as parameter, you are not passing the whole event. So again, there is no need to select currentTarge.value, just the e variable will get you the input value.

src/components/AddFood.js
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        handleFormSubmit = (event) => {
+            event.preventDefault() // always do this with submit events
+            let foodName = event.currentTarget.food.value
+            let image = event.currentTarget.image.value
+            let calories = event.currentTarget.calories.value
+            let newfoods = [{name:foodName, calories: calories, image:image, completed: false}, ...this.state.foods]
+            this.setState({
+              foods: newfoods,
```
IH-WebDev-TA-Remote, line 27: up to this point the code is perfect, but then when you get the new food you are just saving that into the state. The idea would be to use the handleAddfood you are passing through props and invoke it with the food element. That way the function will run in App.js and update the elements with the new food.

src/App.js
```diff
@@ -1,26 +1,80 @@
...
+    foods: foods,
+    filtered: foods,
+    TodaysFoods: { },
+    };
+
+}
+
+  handleAddfood = (event) => {
+    event.preventDefault() // always do this with submit events
+    let foodName = event.currentTarget.food.value
```
IH-WebDev-TA-Remote, line 26: same as the other method, the idea is that in the AddFood component you already have access to the food that is going to be added, so there is no need to access the currentTarge.food.value.

IH-WebDev-TA-Remote:
There are several issues preventing the search bar and add food functionality to run. Check the individual notes for more info. Special attention to how the event is passed when calling functions. Also always using PascalCasing when creating or importing components.

PR #682, opened 2020-08-21
[WDFT-REM-2007] --- LAB-react-ironNutrition ---    Joel Roca Martinez

JoelRocaMartinez:
Sorry for the dely. Manish just reminded me today that I didn´t submit this LAB. I just left it like I had it. I only managed to finish until Iteration 2 the day we had to do it and later, instead of that lab,  I was working on the bonus lab called "lab-react-training" as a recomendation from Manish. I think it´s just fair to submit like I had it :)

### DIFF
--- a/src/components/FoodBox.js
+++ b/src/components/FoodBox.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import foods from '../foods.json';
+import FoodDetails from "./FoodDetails";
+import 'bulma/css/bulma.css';
+
+
+
+
+class FoodBox extends React.Component {
+
+    constructor(props) {
+      super(props)
+      this.state = {
+          listOfFood: foods 
+      } 
+    }
+    render() {
+
+
+
+      return (
+        <div>
+        {
+            this.state.listOfFood.map((food, i) => {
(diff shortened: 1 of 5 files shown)


src/components/FoodBox.js
```diff
@@ -0,0 +1,36 @@
...
+
+    constructor(props) {
+      super(props)
+      this.state = {
+          listOfFood: foods 
+      } 
+    }
+    render() {
+
+
```
IH-WebDev-TA-Remote, line 19: watch spacing here

src/components/FoodDetails.js
```diff
@@ -0,0 +1,46 @@
+import React from "react"
+
+
+
+function FoodDetails(props) {
+
+/*
+  props = {
```
IH-WebDev-TA-Remote, line 8: to work around a state like this you will need to use the hooks syntax or change it into a class component.

IH-WebDev-TA-Remote:
Quite a lot missing but the first two iterations are good and code is easy to follow. Seems you managed to understand better props and state with the other two labs so right now I would suggest to continue focusing on your projects.

PR #697, opened 2020-09-25
[REMOTE-WDPT-0602] – Willem Prins

willemprins93:
Done, except for the bonus iterations.


willemprins93:
Finished all iterations. If possible, would like an overview of the last iteration to see if my solution is good practice or not.

PR #704, opened 2020-09-26
[REMOTE-WDPT-0602] -Cristina Matamoro

cris-developer:
Done till iteration 4. I have doubts when using bind. Could be shown to us some examples about it?Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,13874 @@
+{
+  "name": "starter-code",
+  "version": "0.1.0",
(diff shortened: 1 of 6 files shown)


itstheandre:
Hey Cristina, you no longer have to use the `.bind` . Or the constructor even. As long as you do a state directly:

```
class FancyComponent extends React.Component {
state = {
something: true
}
```

As long as you use arrow functions in the methods you dont need to use thing `.bind`

PR #752, opened 2020-10-28
[REM WDFT 2009] Marta Obregon

MartaObregon:
Iteration 5 is missing! Didn´t have the time today to finish, sorry!

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,74 @@
-import React from 'react';
-import logo from './logo.svg';
+import React, { Component } from 'react'
+import 'bootstrap/dist/css/bootstrap.css'
 import './App.css';
+import 'bulma/css/bulma.css';
+import foods from './foods.json';
+import FoodBox from './components/FoodBox'
+import AddItem from './components/AddItem'
+import {Button} from 'react-bootstrap'
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+class App extends Component {
+
+  state = {
+    foodsList: foods,
+    showForm: false,
+    filteredList: foods,
+  }
+
+  handleAdd = (event) =>{
+    event.preventDefault()
+    console.log(event.target)
+    let newFood = {
+      id: this.state.foodsList.length +1,
+      name: event.target.itemName.value,
+      imageUrl: event.target.itemImage.value,
+      calories: event.target.itemCalories.value,
+      quantity: 0
+    }
+
+    this.setState({
+      foodsList: [...this.state.foodsList, newFood],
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,26 +1,74 @@
...
+    foodsList: foods,
+    showForm: false,
+    filteredList: foods,
+  }
+
+  handleAdd = (event) =>{
+    event.preventDefault()
+    console.log(event.target)
+    let newFood = {
+      id: this.state.foodsList.length +1,
```
IH-WebDev-TA-Remote, line 22: there is no use for this id property so no need to add it. If there was a delete functionality this might create ids tha are the same.

src/App.js
```diff
@@ -1,26 +1,74 @@
...
+      name: event.target.itemName.value,
+      imageUrl: event.target.itemImage.value,
+      calories: event.target.itemCalories.value,
+      quantity: 0
+    }
+
+    this.setState({
+      foodsList: [...this.state.foodsList, newFood],
+      showForm: false,
```
IH-WebDev-TA-Remote, line 31: nice that you hide the form after submitting!

src/App.js
```diff
@@ -1,26 +1,74 @@
...
+  handleShowForm = () =>{
+    this.setState({
+      showForm:true
+    })
+  }
+
+  handleSearch = (event) => {
+    let searchText = event.target.value
+
+    const {foodsList} = this.state
```
IH-WebDev-TA-Remote, line 45: good use of destructuring. In the previous function it would have help too.

src/App.js
```diff
@@ -1,26 +1,74 @@
...
+
+    let filteredList = foodsList.filter((item)=>{
+      return item.name.toLowerCase().includes(searchText.toLowerCase())
```
IH-WebDev-TA-Remote, line 48: good use of toLowerCase to make the search functionality not case sensitive.

src/components/FoodBox.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from 'react'
+import FoodItem from './FoodItem'
+
+export default class FoodBox extends Component {
```
IH-WebDev-TA-Remote, line 4: since there is no state here to manage, this could be a function component. This could even be implemented directly in App.js without the need of a component.

src/components/FoodBox.js
```diff
@@ -0,0 +1,22 @@
...
+    render() {
+
+     const{items} = this.props
+
+        return (
+          <div>
+            {
+                items.map((item)=>{
+                    return <FoodItem
```
IH-WebDev-TA-Remote, line 13: always remember to add keys to any element inside a loop in react. This could be the index of the element taken from the parameter.

src/components/FoodItem.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from 'react'
+
+export default class FoodItem extends Component {
+    render() {
+     const {singleItem} = this.props
```
IH-WebDev-TA-Remote, line 5: instead of destructuring the whole item, you could destructure the specific properties, like image, calories and quantity.

IH-WebDev-TA-Remote:
Good solution, code is clean and easy to read. for iterations 5 and up check the solutions we will send, those are hard but very important to know how they are done since your projects will reply a lot on complex connections like this one. Keep it up!

PR #753, opened 2020-10-28
[ REM 2009 ] Aleksandra Tanasijenko

alekstana:
4,5 iterations done. 
Gonna continue later tonight!!!

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,104 +2,129 @@ import React, { Component } from 'react';
 import './App.css';
 import 'bulma/css/bulma.css';
 import foods from './foods.json';
-import AddFood from './components/AddFood'
-import Foodbox from './components/FoodBox'
+import AddFood from './components/AddFood';
+import Foodbox from './components/FoodBox';
+import Menu from './components/Menu';
+
 
 class App extends Component {
-  
   state = {
     foodList: foods,
     filteredList: foods,
-    showForm: false
-  }
-
-
+    showForm: false,
+    todaysFood:[],
+    allCalories:[]
+  };
 
   handleAdd = (event) => {
-    event.preventDefault()
-    console.log(event.target)
+    event.preventDefault();
+    console.log(event.target);
 
-    let food = event.target
+    let food = event.target;
     let newFood = {
-        name: food.name.value,
-        calories: food.calories.value,
-        quantity: 0,
-        image: food.image.value, 
-    }
+      name: food.name.value,
+      calories: food.calories.value,
+      quantity: 0,
+      image: food.image.value,
+    };
 
     this.setState({
       foodList: [...this.state.foodList, newFood],
       filteredList: [...this.state.filteredList, newFood],
-      showForm: false
-    })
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,26 +1,131 @@
...
+    let newFood = {
+      name: food.name.value,
+      calories: food.calories.value,
+      quantity: 0,
+      image: food.image.value,
+    };
+
+    this.setState({
+      foodList: [...this.state.foodList, newFood],
+      filteredList: [...this.state.filteredList, newFood],
```
IH-WebDev-TA-Remote, line 33: by putting the element first you can see the element being added to the beginning of the list instead of the end.

src/App.js
```diff
@@ -1,26 +1,131 @@
...
+  handleFormShow = () => {
+    this.setState({
+      showForm: true,
+    });
+  };
+
+  handleSearch = (event) => {
+    event.preventDefault();
+    let search = event.target.value;
+    const { foodList } = this.state;
```
IH-WebDev-TA-Remote, line 47: good use of destructure, you could have even do it in the previous function.

src/App.js
```diff
@@ -1,26 +1,131 @@
...
+
+    let searchedList = foodList.filter((food) => {
+      return food.name.toLowerCase().includes(search.toLowerCase());
```
IH-WebDev-TA-Remote, line 50: very nice use of toLowerCase to make the search functionality not case sensitive.

src/App.js
```diff
@@ -1,26 +1,131 @@
...
+            type="text"
+            placeholder="Search"
+          ></input>
+
+          <hr></hr>
+        </div>
+        <div className="columns">
+          <div className="column">
+            {filteredList.map((food) => {
+              return <Foodbox food={food} toSum={this.handleTodaysFood}/>;
```
IH-WebDev-TA-Remote, line 109: remember to add a key to every item that is rendered in a loop in react. This could even be the index that comes from the parameter.

src/App.js
```diff
@@ -1,26 +1,131 @@
...
+            })}
+          </div>
+          <div className="column">
+            <h2 className="subline">Today's foods</h2>
+            <hr></hr>
+            {todaysFood.map((item) => {
+              return <Menu item={item}/>;
+            })}
+            <p> <span className="bold">Total:</span> 
+            {allCalories.reduce((a,b) => {
```
IH-WebDev-TA-Remote, line 119: very nice implementation of the total calories with reduce. nicely done!

src/App.js
IH-WebDev-TA-Remote: toSum might have a different meaning than handleTodaysFood. Try to have similar names for your props than the function or element being passed. You can even use the same name if you want.

IH-WebDev-TA-Remote:
Very good job! and awesome that you finished iteration 5. Also nice styles, code is clean and easy to follow. Check the solution to see the approach for the bonuses. Keep it up!

PR #754, opened 2020-10-28
[REM-2009] Laura Salakari

LauraSalakari:
c:

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,114 @@
-import React from 'react';
-import logo from './logo.svg';
+import React, { Component } from 'react'
+import "bulma/css/bulma.css";
 import './App.css';
+import foods from "./foods.json";
+import FoodList from './components/FoodList';
+import AddFoods from './components/AddFoods';
+import TodaysFoods from './components/TodaysFoods';
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+export default class App extends Component {
+
+  state = {
+    foodList: foods,
+    filteredFoods: foods,
+    todaysFoods: [],
+    showAddForm: false
+  }
+
+  handleShowAddForm = () => {
+    this.setState({
+      showAddForm: true
+    })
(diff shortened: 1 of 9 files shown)


src/App.js
```diff
@@ -1,26 +1,114 @@
...
+    this.setState({
+      foodList: [...this.state.foodList, newFood],
+      filteredFoods: [...this.state.filteredFoods, newFood],
+      showAddForm: false
+    })
+  }
+
+  handleSearch = (e) => {
+    let search = e.target.value;
+    const { foodList } = this.state;
```
IH-WebDev-TA-Remote, line 42: Same as here, you could have do a destructure of the state in the previous function.

src/components/FoodBox.js
```diff
@@ -0,0 +1,39 @@
...
+                            </p>
+                        </div>
+                    </div>
+                    <div className="media-right">
+                        <div className="field has-addons">
+                            <div className="control">
+                                <input id={"form-quantity-input" + item.name} className="input" type="number" defaultValue="1" />
+                            </div>
+                            <div className="control">
+                                <button onClick={() => { this.props.onAddTodaysFood(item.name, document.getElementById("form-quantity-input" + item.name)) }} className="button is-info">
```
IH-WebDev-TA-Remote, line 29: this is a VERY clever way to solve this, but with react you always have access to the state which you use to get and modify information from a component. Using the state will have better performance than accessing the element via DOM.

src/App.js
```diff
@@ -1,26 +1,114 @@
...
+    })
+
+    this.setState({
+      filteredFoods: filtered
+    })
+  }
+
+  handleTodaysFood = (food, quantity) => {
+    let updateList = this.state.foodList.map((elem) => {
+      if (elem.name === food) elem.quantity += Number(quantity.value);
```
IH-WebDev-TA-Remote, line 55: very clever how you managed to set the quantity from within every element and modify the state based on that info. Both here and on the handleDelete.

src/components/FoodList.js
```diff
@@ -0,0 +1,22 @@
...
+import FoodBox from './FoodBox';
+
+export default class FoodList extends Component {
+    render() {
+        const {foods} = this.props;
+
+        return (
+            <div>
+                {
+                    foods.map((item, i) => {
```
IH-WebDev-TA-Remote, line 11: making this a component may not be needed. since right now it is only running the map. Instead you could run the map in App.js and avoid this part. The result will be the same with way less code.

IH-WebDev-TA-Remote:
Very well done! some approaches there are very clever like the total calories and adding/deleting elements from the todays food. Code is easy to follow and clean. Keep up the amazing work!

PR #755, opened 2020-10-28
[REM-2009] Sandra Mendez

scmendez:
Mandatory iterations complete, although the formatting definitely leaves something to be desired. Bonuses not attempted, brain tired.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,24 +1,99 @@
-import React from 'react';
-import logo from './logo.svg';
+import React, {useState} from 'react';
 import './App.css';
 
-function App() {
+import 'bulma/css/bulma.css';
+import foods from './foods.json';
+
+import FoodBox from './components/foodbox/FoodBox'
+import AddFood from './components/addfood/AddFood'
+import Search from './components/search/Search'
+
+const App = () => {
+
+  const [foodList, setFoodList] = useState(foods)
+  const [foodListClient, setFoodListClient] = useState(foods)
+  const [showForm, setShowForm] = useState(false)
+  const [todaysFoods, setTodaysFoods] = useState([])
+  const [totalTodaysCalories, setTotalTodaysCalories] = useState([])
+
+  const handleShowForm = () => {
+    setShowForm(true)
+  }
+
+  const handleNewFoodSubmit = (event) => {
+    event.preventDefault()
+
+    let newFoodName = event.target.name.value
+    let newFoodCalories = event.target.calories.value
+    let newFoodImg = event.target.image.value
+
+    let newFood = {
+      name: newFoodName,
+      calories: newFoodCalories,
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -1,24 +1,99 @@
...
+import Search from './components/search/Search'
+
+const App = () => {
+
+  const [foodList, setFoodList] = useState(foods)
+  const [foodListClient, setFoodListClient] = useState(foods)
+  const [showForm, setShowForm] = useState(false)
+  const [todaysFoods, setTodaysFoods] = useState([])
+  const [totalTodaysCalories, setTotalTodaysCalories] = useState([])
+
```
IH-WebDev-TA-Remote, line 18: amazing that you are going full hooks with these! :)

src/App.js
```diff
@@ -1,24 +1,99 @@
...
+  const handleShowForm = () => {
+    setShowForm(true)
+  }
+
+  const handleNewFoodSubmit = (event) => {
+    event.preventDefault()
+
+    let newFoodName = event.target.name.value
+    let newFoodCalories = event.target.calories.value
+    let newFoodImg = event.target.image.value
```
IH-WebDev-TA-Remote, line 28: to reduce the code a bit you can directly create the obj with the values from the target instead of saving that into variables.

src/App.js
```diff
@@ -1,24 +1,99 @@
...
+    })
+
+    setFoodListClient(searchedList)
+  }
+
+  const handleTodayListAdd = (foodInfo, quantity) => {
+    let todaysFoodName = foodInfo.name
+    let totalCalories = foodInfo.calories * quantity
+
+    let todaysFoodItem = `${quantity} ${todaysFoodName} = ${totalCalories} calories`
```
IH-WebDev-TA-Remote, line 56: this works but it would make more sense that you pass the information as an object with its properties quantity, name and calories. that was you can arrange the information in the JSX better.

src/components/foodbox/FoodBox.js
```diff
@@ -0,0 +1,46 @@
+import React, {useState} from 'react'
+
+const FoodBox = (props) => {
+
+    const [quantity, setQuantity] = useState(0)
+
+    const handleQuantityChange = (event) => {
+        let foodQuantity = event.target.value
+        setQuantity(foodQuantity)
+    }
```
IH-WebDev-TA-Remote, line 10: perfectly handled how you pass the quantity information!

IH-WebDev-TA-Remote:
Very well done! Amazing how you are using hooks so easily. All the mandatory parts are done. Code is clean and easy to read. Check the solution and try to understand the logic using classes (it is good to at least being able to understand how those work) as well as the bonus iterations. Keep up the great work! :)

PR #759, opened 2020-10-28
[REM-2009] Sumiya Ushiro

sumi0820:
Completed all iterations and bonus. However, on iteration 5 and 6, I was not able to come up with better solutions to make codes shorter, and it ended up becoming redundant and messy. Please review it and give me some advice to make it more readable.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,127 @@
 import React from 'react';
-import logo from './logo.svg';
+import FoodBox from './components/FoodBox';
+import AddFood from './components/AddFood';
+import Search from './components/Search';
+
+import foods from './foods.json';
+
+import 'bulma/css/bulma.css';
 import './App.css';
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+class App extends React.Component {
+  state = {
+    cloneFoods: foods,
+    addForm: false,
+    todaysFoods: [],
+    totalCal: 0,
+  };
+
+  // Form handler //
+  handleAdd = (e) => {
+    e.preventDefault();
+    let name = e.target.addFoodName.value;
+    let image = e.target.addFoodImage.value;
+    let calories = e.target.addFoodCalories.value;
+
+    let newFood = {
+      name: name,
+      image: image,
+      calories: calories,
+      quantity: 0,
+    };
+    this.setState({
+      cloneFoods: [...foods, newFood],
+      addForm: false,
+    });
+  };
+
+  handleForm = () => {
+    this.setState({
+      addForm: true,
+    });
+  };
+
+  // Search handler //
+  handleSearch = (e) => {
+    e.preventDefault();
+    const { cloneFoods } = this.state;
+    let searchInput = e.target.value.toLowerCase();
+    console.log(searchInput);
+    let results = foods.filter((food) => {
+      return food.name.toLowerCase().includes(searchInput);
+    });
+    this.setState({
+      cloneFoods: results,
+    });
+  };
+
+  // Today's food handler //
+  handleTodaysFood = (item, qty) => {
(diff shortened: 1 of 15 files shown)


src/App.js
```diff
@@ -1,26 +1,126 @@
...
+    todaysFoods: [],
+    totalCal: 0,
+  };
+
+  // Form handler //
+  handleAdd = (e) => {
+    e.preventDefault();
+    let name = e.target.addFoodName.value;
+    let image = e.target.addFoodImage.value;
+    let calories = e.target.addFoodCalories.value;
```
IH-WebDev-TA-Remote, line 24: you can avoid above 3 lines of code by simply creating the object below directly with the target values

src/App.js
```diff
@@ -1,26 +1,126 @@
...
+
+    let newFood = {
+      name: name,
+      image: image,
+      calories: calories,
+      quantity: 0,
+    };
+    this.setState({
+      cloneFoods: [...foods, newFood],
```
IH-WebDev-TA-Remote, line 33: **foods** is the original object, so when you try to add a new element you will never see the previous ones added. So instead of **...foods** try using the current state: **...this.state.cloneFoods**

src/App.js
```diff
@@ -1,26 +1,126 @@
...
+      addForm: true,
+    });
+  };
+
+  // Search handler //
+  handleSearch = (e) => {
+    e.preventDefault();
+    const { cloneFoods } = this.state;
+    let searchInput = e.target.value.toLowerCase();
+    let results = cloneFoods.filter((food) => {
```
IH-WebDev-TA-Remote, line 49: the cloneFoods state will continue to reduce everytime you change the search input, meaning it won't include the original data and if you try to delete the input in won't find any matches from the original data. 
To solve this quick you can use the original data variable **foods.filter** instead of the state **cloneFoods.filter**
A better solution though, would be to have a separate state that holds the original data and will never be altered, that way you can access this info from the state when you need all the info and not something that was changed along the way.

src/components/FoodItem.js
```diff
@@ -0,0 +1,63 @@
...
+
+class FoodItem extends Component {
+  state = {
+    qty: 0,
+  };
+
+  handleQty = (e) => {
+    this.setState({
+      qty: e.target.value,
+    });
```
IH-WebDev-TA-Remote, line 12: perfect how you handled having a state for information that will be changed inside the component.

src/App.js
```diff
@@ -1,26 +1,126 @@
...
+
+      let newFoods = todaysFoods.map((food) => {
+        if (food.name == item.name) {
+          food = {
+            name: food.name,
+            calories:
+              newFood.calories * (parseInt(food.quantity) + parseInt(qty)),
+            quantity: parseInt(food.quantity) + parseInt(qty),
+          };
+          duplicate = true;
```
IH-WebDev-TA-Remote, line 76: It is actually a good approach using the duplicate variable. Another solution would be to put the setState with the different conditions here, however the duplicate variable is actually a cook approach as well. well done.

src/App.js
```diff
@@ -1,26 +1,126 @@
...
+      });
+      duplicate = false;
+    }
+  };
+
+  handleDelete = (itemName) => {
+    const { todaysFoods } = this.state;
+    let filteredFoods = todaysFoods.filter((food) => {
+      return itemName !== food.name;
+    });
```
IH-WebDev-TA-Remote, line 94: to update the calories amount after you delete an element you can run a new .reduce on all the elements from todaysFood and then update the totalCal state.

src/components/TodaysFood.js
```diff
@@ -0,0 +1,26 @@
...
+          <p>{foodInfo.quantity} qty</p>
+          <p>{foodInfo.name}</p>
+          <p>{foodInfo.calories} cal</p>
+        </div>
+        <button
+          onClick={() => {
+            onDelete(foodInfo.name);
+          }}
+        >
+          <a class="delete"></a>
```
IH-WebDev-TA-Remote, line 21: remember to change all your **class** to **className**

IH-WebDev-TA-Remote:
Very well done, love the styles you added! However the functionalities for adding new foods and search were a little broken, check the feedback to see how to fix and improve them. As per your question, the code for iteration 6 looks very good, you will definitely need to run a map to see if the foods are repeated. You could try to reduce syntax but the logic is needed. You can check the solution shared if you want to see a different approach but there is still a lot of code there.

Overall though, very good job. Keep it up!

PR #761, opened 2020-10-29
[REM 2009] Mohammed Imran

14imran:


src/App.js
```diff
@@ -1,25 +1,104 @@
...
 
 function App() {
+  const [foods, setFoods] = useState([]);
+  const [searchedFoods, setSearchFoods] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [isForm, setIsForm] = useState(true);
+  const [query, setQuery] = useState('');
+
+  const [addingFood, setAddingFood] = useState([]);
+  const [addedFood, setAddedFood] = useState([]);
```
IH-WebDev-TA-Remote, line 18: very good getting use to hooks! awesome

src/App.js
```diff
@@ -1,25 +1,104 @@
...
+
+  useEffect(() => {
+    setFoods(foodsData);
+    setTimeout(() => {
```
IH-WebDev-TA-Remote, line 22: this might be needed when fetching information from an external source, but in this case the setTimout is not required. It will run perfectly without it.

src/App.js
```diff
@@ -1,25 +1,104 @@
...
+    }, [1000]);
+  }, []);
+
+  let handleAdd = (event) => {
+    event.preventDefault();
+    let name = event.target.name.value;
+    let calories = event.target.calories.value;
+    let image = event.target.image.value;
+    setFoods([...foods, { name, calories, image }]);
+    setIsForm(true);
```
IH-WebDev-TA-Remote, line 33: very nice reducing the syntax of your code!

src/App.js
```diff
@@ -1,25 +1,104 @@
...
+      )
+    );
+  };
+
+  const handleAddButton = (foodDetails, value) => {
+    console.log('hellofrom App', value);
+    let name = foodDetails.name;
+    let calories = foodDetails.calories * value;
+
+    let total = `${value} ${name} = ${calories} calories   `;
```
IH-WebDev-TA-Remote, line 54: instead of creating a string, you may want to create objects and add those to the state. That way you can easily modify the elements and change how each element is rendered in the JSX.

src/App.js
```diff
@@ -1,25 +1,104 @@
...
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
+    <>
+      {isLoading ? (
+        <h1>loading </h1>
+      ) : (
```
IH-WebDev-TA-Remote, line 66: good use of the isLoading state and the lifecycle methods

src/App.js
```diff
@@ -1,25 +1,104 @@
...
+        <div className="App">
+          {isForm ? (
+            <>
+              <button onClick={hanleShowForm}>show</button>
```
IH-WebDev-TA-Remote, line 70: very clever using a state to render the form

src/App.js
```diff
@@ -1,25 +1,104 @@
...
+              <input
```
IH-WebDev-TA-Remote, line 71: this searchbar could be a component itself, right now it might not be needed elsewhere but in other situations it could be reused.

src/App.js
```diff
@@ -1,25 +1,104 @@
...
+              <div style={{ marginLeft: '60%' }}>
+                Today's Food
+                <p>{addingFood}</p>
+              </div>
+            </>
+          ) : (
+            <AddFood onAdd={handleAdd} />
+          )}
+
+          {query == ''
```
IH-WebDev-TA-Remote, line 86: instead of this ternary, you could update the searchedFoods in the useEffect method and it will show the initial foods even without using the search bar, which is a better approach.

src/component/FoodBox.js
```diff
@@ -0,0 +1,60 @@
+import React, { useState } from 'react';
+
+export default function FoodBox(props) {
+  const { name, calories, image, quantity } = props.food;
+  const [value, setValue] = useState(0);
```
IH-WebDev-TA-Remote, line 5: perfect approach using a state for the quantity

IH-WebDev-TA-Remote:
Very good solution, and all using hooks and lifecycle methods, very well done. For the last bonus iterations check the solutions provided and see if you can translate the approaches to function components and hooks. Keep it up!

PR #762, opened 2020-10-29
rem 2009 katie stamey

kmstamey:


kmstamey:
this was very difficult for me. i've watched most of the tuesday lecture but now I want to rewatch the video from Wednesday just to be sure. This was a lot of back and forth with the exercise from class and my boyfriend's help. I feel like i'm ~45% understanding but looking forward to seeing it done in class

src/App.js
```diff
@@ -1,26 +1,128 @@
...
+  handleShowForm = () => {
+    this.setState({
+      formShow: true
+    })
+  }
+
+  handleSearch = (event) => {
+    let searchName = event.currentTarget.value
+    let cloneFoods = this.state.foods.filter((item) => {
+      return item.name.startsWith(searchName)
```
IH-WebDev-TA-Remote, line 51: instead of startsWith you can use the include methods. That will search inside any part of the strings instead of specifically how the string starts. Also if you apply toLowerCase to both the searchName and the item.name you can improve the functionality of the search to be not case sensitive.

src/App.js
```diff
@@ -1,26 +1,128 @@
...
+          <input 
+            type="text" 
+            onChange={this.handleSearch}
+            placeholder="Search">
+          </input>
+
+         
+        
+        {
+          this.state.foods.map((food, index) =>{
```
IH-WebDev-TA-Remote, line 98: this is what was breaking your search bar. When you search you update the filteredFoods state but here you always show the original foods state. try changing that to be **this.state.filteredFoods.map** and it will work

src/App.js
```diff
@@ -1,26 +1,128 @@
...
+    })
+
+    this.setState({
+      filteredFoods: cloneFoods
+    })
+  }
+
+  //ITERATION 5
+
+  handleToday = (event) =>{
```
IH-WebDev-TA-Remote, line 61: the code of this function looks good, the part missing is how you were to pass this whole information from the FoodBox component.

IH-WebDev-TA-Remote:
Good job on the first iterations but definitely there was a lot left to be done. When possible check the solutions provided to see how to approach these solutions and try to go back to it and do it yourself with what you have learned, these will be essential logics to be applied on your project. 

Finally remember to ask for help when you get stuck during any iteration and let us know if you need a hand understanding anything in the solutions :) Keep it up!

PR #792, opened 2020-11-25
[FTAMS102020] Nina van den Braak

ninsajala:
Still want to look at the bonus iterations

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -8,13 +8,34 @@ import AddFoodForm from './addfood';
 function App() {
   const [foods, setFoods] = useState(foodList);
   const [showForm, setShowForm] = useState(false);
+  const [searchInput, setSearchInput] = useState('');
 
   const toggleForm = () => setShowForm(!showForm);
 
   const addFoodHandler = (newFood) => {
     const foodCopy = [...foods];
+    foodList.push(newFood);
     foodCopy.push(newFood);
     setFoods(foodCopy);
+  };
+
+  const handleSearchInput = (event) => {
+    event.preventDefault();
+    setSearchInput(event.target.value);
+  };
+
+  const searchFoods = (event) => {
+    event.preventDefault();
+    let foodCopy = [...foods];
+    if (searchInput === '') {
+      setFoods(foodList)
+    } else {
+      setFoods(
+        foodCopy.filter((item) =>
+          item.name.toUpperCase().includes(searchInput.toUpperCase())
+        )
+      );
+    }
   };
 
   return (
@@ -22,6 +43,13 @@ function App() {
       <h1>IronNutrition</h1>
       <div className="foodWrapper">
         <div className="foodList">
+          <form className="search" onSubmit={searchFoods}>
+            <input
+              type="text"
+              placeholder="Search for food"
+              onChange={handleSearchInput}
+            />
+          </form>
           {foods.map((item, index) => (
             <Foodbox key={index} food={item} />
           ))}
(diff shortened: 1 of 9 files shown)


src/App.js
```diff
@@ -1,24 +1,96 @@
...
+                foodTotal={totalEaten}
+                deleteHandler={deleteEatenItem}
+              />
+            )}
+          </div>
+          <div className="addFood">
+            {showForm ? (
+              <AddFoodForm addFood={addFoodHandler} formToggler={toggleForm} />
+            ) : (
+              <button class="button is-primary" onClick={() => toggleForm()}>
```
LloydC, line 87: You should have used `className` instead of `class here`.

src/addfood.js
```diff
@@ -0,0 +1,83 @@
...
+    setAllValues({ ...allValues, [name]: value });
+  };
+
+  const handleFormSubmission = (event) => {
+    event.preventDefault();
+    props.addFood(allValues);
+    props.formToggler();
+    setAllValues(initialState);
+  };
+
```
LloydC, line 24: Same comment as above, all your classes should be changed to `className` here.

src/search.js
```diff
@@ -0,0 +1,35 @@
...
+  const handleSearchInput = (event) => {
+    event.preventDefault();
+    setSearchInput(event.target.value);
+  };
+
+  const handleSubmit = (event) => {
+    event.preventDefault();
+    props.searchFood(searchInput);
+  };
+
```
LloydC, line 15: Same comment as above, all your classes should be changed to `className` here.

src/search.js
```diff
@@ -0,0 +1,35 @@
...
+        <div class="control">
+          <input
+            class="input"
+            type="text"
+            placeholder="Search for food"
+            onChange={handleSearchInput}
+          />
+        </div>
+        <div class="control">
+          <p class="button is-info">Search</p>
```
LloydC, line 28: It would make more sense to use a `<button type="submit">` instead of `<p>` so you can submit the form.

src/search.js
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react';
+
+function Search(props) {
+  const [searchInput, setSearchInput] = useState('');
+
+  const handleSearchInput = (event) => {
+    event.preventDefault();
```
LloydC, line 7: Since you're passing this function to an `<input onChange>`, you don't need to use `event.preventDefault()` in this case.

Generally speaking, you'll only use it in React to prevent a form from being submitted.

src/foodbox.js
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from 'react';
+
+function Foodbox(props) {
+  const [quantity, setQuantity] = useState(props.food.quantity);
+
+  const handleQuantity = (event) => {
+    event.preventDefault();
```
LloydC, line 7: Again, you don't need `event.preventDefault()` here since you're not working with a form element.

src/foodbox.js
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  let foodForTotal = {
+    name: props.food.name,
+    quantity,
+    calories: quantity * props.food.calories,
+  };
+
+  const addToTotal = (event) => {
+    event.preventDefault();
```
LloydC, line 18: Same comment as above.

ninsajala: Thanks for the feedback Lloyd. The classes were a bit carelessly copied from Bulma I have to admit.. 
All is resolved now!

PR #809, opened 2020-11-27
[FTAMS102020] Emma Northridge

emmanorthridge:
Used  code from the code along, but re-wrote everything - helped me solidify more understanding of the topic!

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,21 @@
-import React from 'react';
-import logo from './logo.svg';
 import './App.css';
+import React, { Component } from 'react';
+import 'bulma/css/bulma.css';
+import foodList from './foods.json';
+import FoodBox from './Components/FoodBox';
 
-function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -1,26 +1,97 @@
...
+    stateCopy.filteredFoodListState = filteredFoodList;
+    stateCopy.searching = true;
+
+    this.setState(stateCopy);
+  };
+
+  handleAddFoodToTodaysList = (foodObject) => {
+    const stateCopy = { ...this.state };
+    foodObject.calories *= foodObject.quantity;
+    stateCopy.todaysFoodList.push(foodObject);
```
wax-bear, line 46: One question to consider here is whether you would like to have instances of the same item listed separately, or whether you would like to increment the calories and quantity values of the item that already exists in the list, in case the item you are pushing already exists in the todaysFoodList.

You could do something like:

```
let foundFood = stateCopy.todaysFoodList.find(el => el.name === foodObject.name);

foodObject.calories *= food.quantity;

if (foundFood) {
    foundFood.quantity += food.quantity;
    foundFood.calories += food.calories;
} else {
    today.push(food);
}

this.setState({
    today: stateCopy.todaysFoodList
});
```

src/Components/FoodForm.js
```diff
@@ -0,0 +1,69 @@
+import React, { useState } from 'react';
+
+const initialState = {
+  name: '',
+  calories: 0,
+  image: '',
+  quantity: 0,
```
wax-bear, line 7: In Foodbox, a quantity of 1 is set to the state.
Will setting a quantity here and passing it to foodBox have an effect?

wax-bear (commented):
Well done Emma!
Please check my comments for potential improvements.
Right now, no image can be added as far as I can see. This could also be implemented, if you happen to find the time for that.

PR #811, opened 2020-11-27
[FT-AMS-102020] Elia Chiarucci

eliachiarucci:


src/components/FoodList/FoodList.jsx
```diff
@@ -0,0 +1,41 @@
+import React, { useState } from 'react';
+
+const FoodList = (props) => {
+  const [quantity, setQuantity] = useState(1);
+
+  const addSelectedFood = (food) => {
+    //console.log(food);
+    food.quantity += parseInt(quantity);
```
wax-bear, line 8: Perhaps change this to the following: 

```
    if (food.quantity) {
      food.quantity += quantity;
    } else {
      food.quantity = quantity;
    }
```

If a new food item is added, this ensures the quantity will be passed to the calories list. Without it, I encounter a "NaN" in the calories.

wax-bear (commented):
Good stuff Elia!

Perhaps you are still working on this lab, I will nonetheless leave some comments:
1. The addition of items to the calories list is not always flawless. For example, when adding bigger numbers, these numbers are concatenated with the previous number, rather than the sum being rendered.
2. As far as I can see I cannot add an image yet.
3. A "class" instead of "className" is used, indicated by an error being logged in the console.
4. Your code looks well formatted. Good stuff!

Please also check my comments and reach out if you have a question.

eliachiarucci:
Hi Thomas,
I always forget to remove the comments... 😄 
Anyway, I think It's fixed now, the new food was inserted in the state with "quantity: NaN" and string calories, I parsed those in the handle function.
As for the image, I tried and I think it works as it should by pasting a link, but let me know.

wax-bear:
Great stuff Elia!
Indeed, the image addition works by submitting a link. Thanks for notifying me on that. :) 
Happy coding!

PR #813, opened 2020-11-28
[FTAMS102020] Alejandro Pliakas Benitez 

apliakas:


wax-bear (commented):
Good stuff Alex!

A few points of attention:
- Eslint throws errors while compiling the code, which is visible in the console upon npm starting the project. Please fix these errors.
- When submitting no image, I see a "missing image" icon in the list. Perhaps this can be replaced by a standard image.
- Please mind the styling of the code.
- State management is going well. Way to go!

src/App.js
```diff
@@ -1,24 +1,56 @@
...
-          Learn React
-        </a>
-      </header>
+    <div className='AppDiv'>
+      <button onClick={handleForm} className='addButton'>Add new food!</button>
+      {formState && <FoodForm handleForm={handleForm} addFood={addFood}/>}
+      <SearchBar foodFilter={foodFilter} />
+      <div className='body'>
+        <div className='foodBoxes'>
+          {(foodSearch.length === 0 ? foodState : foodSearch).map((foodObject, index) => (
```
wax-bear, line 48: You could even consider getting the foods always from the same array. If the search input is "", the .includes in the foodFilter will return true.

PR #838, opened 2020-12-02
[WDFS-BCN]SERGIO CABRERA CAPINETI

capineti:


src/App.js
```diff
@@ -1,26 +1,139 @@
...
+state={
+  foods: foods,
+  showForm:false,
+
+  name: "",
+  calories: 0,
+  image: "",
+  quantity: 0
+
+
```
tawebbcn, line 19: Take care in adding too many spaces, keep your code clean also in its structure to make it more appealing and clean.

PR #1220, opened 2022-05-11
iteration 1 and 2 done

lukecin:


src/App.js
```diff
@@ -1,24 +1,24 @@
...
+import React, { useState } from 'react';
 import './App.css';
+import 'bulma/css/bulma.css';
+import foods from './foods.json';
+import FoodBox from './components/FoodBox';
+import AddFood from './components/AddFood';
 
 function App() {
+
+  const [foodState, setFoods] = useState(foods);
```
TA-Remote, line 10: Just as convention, be consistent with name for variable and function in useState

either
const [foodsList, setFoodsList] = useState(foods)

or

import foodsList from './foods.json';
const [foods, setFoods] = useState(foodsList);

PR #1221, opened 2022-05-11
[WD REM FT MARS 2022] - Thibaut Lartigue

thibautDevWorld:
Done for today
was tedious but I found some features by myself, it's already 10.23pm. I'll take some rest


src/component/AddFood.js
```diff
@@ -0,0 +1,43 @@
...
+
+        const newFood = {
+            name: name,
+            calories: calories,
+            image: image
+        }
+
+        props.callbackCreateFood(newFood)
+
+
```
TA-Remote, line 24: After we submit the form -> we want to clear the input fields. Better user experience.

You can reset the state to empty fields.  After you update newFood, we can reset the form on line23 like this

     setName('');
     setCalories('');
     setImage('');

PR #1224, opened 2022-05-11
WDFT Remote Mar22-Ha Thu TRAN

HaThu6989:


src/components/FoodBox.js
```diff
@@ -0,0 +1,41 @@
...
+function FoodBox(props) {
+  return (
+    <>
+      {props.foods.map((elem) => {
+        return (
+          <div className="box">
+            <article className="media">
+              <div className="media-left">
+                <figure className="image is-64x64">
+                  <img src={elem.image} />
```
TA-Remote, line 12: remember to add alt key:
```
<img src={elem.image} alt={elem.name}/>
```

src/App.js
```diff
@@ -1,24 +1,42 @@
...
+    setFoodsList((prevFood) => {
+      return [newFood, ...prevFood];
+    });
+  }
+
+  function searchFood(foodToSearch) {
+    setFoodsList((prevFood) => {
+      const newFoodList =
+        foodToSearch === ''
+          ? foodsList
```
TA-Remote, line 27: If you want your whole list to be shown again if searchinput is empty, you have to call your function again, so instead of only 'foodsList' :
```
? setFoodsList(foodsArr);
```

PR #1226, opened 2022-05-11
WDFT2022 PaulineS

Paulinettt:


src/App.js
```diff
@@ -1,24 +1,61 @@
...
+    <Search searchFood={searchFood} />  
+    <br/>
+
+    <button className='button-hide' onClick={hideForm}>{showForm ? "Hide form" : "Add new food"}</button>
+    {showForm && <FoodForm createFood= {createNewFood}/>} 
+    <br/>
+    <br/>
+    <br/>
+   
+    {foods.map((element, index) => {
```
TA-Remote, line 48: Instead of displayig original array, we want to always display filtered array (if there is no search input -> we still have initial value from json)

  {filteredFoods.map((element, index) => {}

src/App.js
```diff
@@ -1,24 +1,61 @@
...
+    setFoods((prevFood)=>{
+      return [newFood, ...prevFood];
+    });
+  }
+
+  const searchFood = (search) => {
+    const searchResults = foods.filter((food) => food.name.includes(search))
+    console.log(searchResults)
+    setFoods(searchResults);
+  }
```
TA-Remote, line 27: Great solution on search functionality! You almost got it!

We only need to make a little tweak 
- add another state , to hold filteredArray and set initial value to original array
```
  const [filteredFoods, setFilteredFoods] = useState(foods);
```

- In searchFood function, instead of updating food , we want to update the value of filtered array 
```
 const searchFood = (search) => {
    const searchResults = foods.filter((food) => food.name.includes(search))
    console.log(searchResults)
    setFilteredFoods(searchResults);
  }
```

- initially, it's value is the same original array from json / or database
- when you filter once (have letters in search input field), we filter teh ORIGINAL array, and set new value of filtered array.
- the second time you add letters to search, or remove letters -> we AGAIN filter from the ORIGINAL array. Because if were filtering from the filtered array furthermore (that's what you have at the moment, we are missign a lot of foods right?)


One last note, when you filter -> we need to make sure it's case insecitive - better user experience.
```
    const searchResults = foods.filter((food) => food.name.toLowerCase().includes(search.toLowerCase()))
```

TA-Remote (commented):
Excellent Pauline, keep it up!

PR #1227, opened 2022-05-11
Marija - WDFT March 2022 Remote

marija-strah:


src/components/Add.js
```diff
@@ -4,29 +4,56 @@ function Add() {
...
+        setImage(e.target.value);
+        };
+
     return (
         <div>
         <form>
             <button onClick={showMyForm}>Add new food</button>
         </form>
         {showForm && (
             <form>
```
TA-Remote, line 48: One very important step with all the forms in react -> you have to handle the submission, do something with the result right? So for example:

```
 <form onSubmit={handleSubmit}>
```

```
  const handleSubmit = (e) => {
    e.preventDefault();

// at this point you have updated the state already
    const newFood = { foodName, foodCalories,  foodImage }; 

// here we call the function of the paretn component and sent new object to be added to original array 
// you need to create that function addFood in App.js
    props.addFood(newFood);

// here we reset the form input fields to empty values
    setName('');
    setCalories('');
    setImage('');
  };
```

PR #1228, opened 2022-05-11
[FTRMT032022] Oscar Gonzalez

oscargonzalezdev:
- some initial problems with the repository.
- too tired and late to continue :/


src/components/Search.js
```diff
@@ -0,0 +1,30 @@
...
+
+    const filter = (searchedFood) => {
+        if (!searchedFood) {
+            props.setFoods(props.foods)
+        } else {
+            props.setFoods((prevValue) => {
+                return prevValue.filter(food => food.name.toLowerCase().includes(searchedFood.toLowerCase()));
+            });
+        }
+    }
```
TA-Remote, line 14: If you write your filter (search) function in this component, you must have a state that hold the value of originla array -> otherwise we have no access to it.

Better approach would be to store this function in app.js -> and pass it down to Search.js  (call it from search with the searc input value)

src/App.js
```diff
@@ -3,6 +3,7 @@ import './App.css';
 import FoodBox from './components/FoodBox';
 import foodsArr from './foods.json';
 import AddFood from "./components/AddFood";
+import Search from "./components/Search";
 
 
 function App() {
```
TA-Remote, line 9: In App.js you should have another state , to hold filteredArray and set initial value to original array -> that's the array you want to iterate over to display in broswer & send it to you child component to filter
```
  const [filteredFoods, setFilteredFoods] = useState(foods);
```

initially, it's value is the same original array from json / or database
when you filter once (have letters in search input field), we filter teh ORIGINAL array, and set new value of filtered array.
the second time you add letters to search, or remove letters -> we AGAIN filter from the ORIGINAL array. Because if were filtering from the filtered array furthermore (that's what you have at the moment, we are missign a lot of foods right?)

TA-Remote (commented):
Good work Oscar, try to use the tips and fix search functionality.

If still makes no sense - let me know and i'll explain it differently.
Karina

PR #1232, opened 2022-05-12
WDFT REMOOTE MARCH22 Jide

JeanDenisD:
Iteration 0-4 done


src/App.js
```diff
@@ -4,16 +4,28 @@ import 'bulma/css/bulma.css';
...
 function App() {
 
   const [foodList, setFoodList] = useState(foods);
+  const [ food,setFoodArr] = useState(foods);
 
-  const handleAddFood = foodObject => setFoodList([...foodList,foodObject])
+  const handleAddFood = foodObject => setFoodList([...foodList,foodObject]);
+
+  const searchFoodList = foodString =>{
+    const newFoodList = (foodString === ''? food: foodList.filter(food=>food.name.toLocaleLowerCase().includes(foodString)))
```
TA-Remote: This is not exactly correct logic at the moment or at least not what we want :) If I start deleting letters from input I am not rendering anything. Currently only when i delete all input - I reset my original array, but we want to be rendering updated array with each letter

- About search functionality:

In App.js you should have two states. One for original array from db and one for filteredArray and set initial value to original array -> that's the array you want to iterate over to display in broswer & send it to you child component to filter

initially, it's value is the same original array from json / or database
when you filter once (have letters in search input field), we filter teh ORIGINAL array, and set new value of filtered array.
the second time you add letters to search, or remove letters -> we AGAIN filter from the ORIGINAL array. Because if were filtering from the filtered array furthermore (that's what you have at the moment, we are missing a lot of foods right?)


- Another little note, you should have `toLowerCase() `on search input as well to make search more accurate

TA-Remote (commented):
Great work Jide!

Left note on search functionality

PR #1234, opened 2022-05-12
WDFT MAR22 Harry

hmarchbank:
solved 1-4


TA-Remote (commented):
Great Harry, keep it up!

The reason why you have same quantity input changing for each food item -> is because you hold all of your foods inside Food component. 

Better approach would be to have a component for a Single food item -> in this case you should map inside app.js, and in each iteration render food component. 

That way Your state will independently belong to each food component, while currently you have one quantity state for your food array

Fyi I uploaded lab solution to student portal, in case you want to check how to solve iteration5

PR #1235, opened 2022-05-12
WDFTMarch 2022 Valentin Paul

Valentin-Paul:
searchbar is not working correctly. when I delete the search it doesnt show all dishes again.
and had some issues with the parentheses/brakets in iteration 5 when I tried to fix the list on the right side.. so I just left it for the moment, so that there's at least a bit of functionability


src/App.js
```diff
@@ -1,24 +1,79 @@
...
+  const searchFood = (search)=>{
+    setSearch(search)
+    setFood((prevFood)=>{
+     const newArr = prevFood.filter((name)=>{
+      return name.name.includes(search)
+      })
+      const result = [...newArr];
+      return result;
+    })
+  }
```
TA-Remote, line 61: The way we structured our code now, we are sending 'filteredFood' variable to FoodBox, so this is the state we want to update when searching (filtering). 

However, we Always need to filter Original array --> foods. If you think about it, when you start typing smth in search input -> we already have filtered array, correct? If I then add another letter or remove one -> with your current code i would be filtering again on already  Filtered array.

So the idea is that we always read seach input value -> but filter original array 

```
  const searchFood = (search)=>{
    console.log(search)
    setFilteredFood(food.filter((name) => name.name.includes(search)))
}
```

TA-Remote (commented):
Great work Valentin, please try to apply the notes I left for you - even if it's on the weekend :) 

If something does not make sense, please tell me and I can explain again differently!
Karina

PR #1255, opened 2022-07-27
[RMFT June 2022] Alex's lab

Virgule09:
I prefer to take a rest rather than tackling the bonuses :).
I did conditional rendering in the codealong so I believe this should not be too much of a problem.


src/App.js
```diff
@@ -1,25 +1,43 @@
...
+
 
 function App() {
+
+  const [food, setFood] = useState(foods);
+
+  const deleteFood = (foodName) => {
+    setFood((prevFoodArr) => {
+        const newList = prevFoodArr.filter((element) => {
+            return element.id !== foodName;
```
TA-Remote, line 16: `return element.id !== foodName;` I think that should be `element.name`

TA-Remote (commented):
- Great work 
  - I assuming that the mistake in the delete was due the fatigue you related in the comment
  - The other "problem" is the search that Luis addressed in the beginning of the lecture :)

src/App.js
Virgule09: Waow I was really tired...
I actually fixed this, but I committed before even testing the delete function...
Then I realised what the mistake was, fixed it. But forgot to commit again and push again...
So in conclusion, yes it was important for me to take a rest yesterday :)

PR #1256, opened 2022-07-27
Zubair pull request 

zubairmalek:
somehow done


src/components/Search.js
```diff
@@ -0,0 +1,19 @@
...
+      <form action="/" method="get">
+        <input
+          type="text"
+          value={props.searchQuery}
+          onInput={(e) => props.setSearchQuery(e.target.value)}
+          id="search-bar"
+          placeholder="Search for food"
+          name="s"
+        />
+        <button type="submit">Search</button>
```
TA-Remote, line 13: We do not want to have button in this case as it refreshes the browser and your newly created itenms disappear

TA-Remote (commented):
great work, interesting solution for the search functionality!  I suggest you also look at the proposed solution for this lab , it's uploaded on student portal. See alternative approach there :)

PR #1263, opened 2022-07-27
[wdftJun22]_lab_nutrition_Christoph_Bochniak_till_It.5

Christoph86:


src/Components/AddFoodForm.js
```diff
@@ -0,0 +1,49 @@
...
+
+    const handleSubmit = (e) => {
+        e.preventDefault();
+
+        const newFood = {
+            name: name,
+            image: image,
+            calories: calories,
+            servings: servings,
+        };
```
TA-Remote, line 18: You could have used the syntax below too (as properties and variables have the same name): 
```
 const newFood = {
            name,
            image,
            calories,
            servings,
        };
```

src/Components/SearchBar.js
```diff
@@ -0,0 +1,28 @@
...
+
+    const [search, setSearch] = useState("");
+
+    const handleSearch = (e) => {
+        e.preventDefault();
+
+        props.setFood(() => {
+            // don't use prevFood, always the whole list [...foods] to filter
+            // avoid just 'e' for the element, "e" have full scope in this function(handleSearch(e)) (e is the event!)
+            return [...foods].filter(element => element.name.includes(e.target.value));
```
TA-Remote, line 14: `return [...foods].filter(element => element.name.includes(e.target.value)); ` That way you can't search for the foods added. If you used the `foodsArr` you would have a problem as you lose foods from the list after one searching (just to keep in mind)

TA-Remote (commented):
- Great work!!
 - If you have time try to implement the changes so you can add new foods and search through all of them 
 - Implement the delete button ^^

PR #1264, opened 2022-07-27
done Lucas Borchardt 

LucasBorchardt:
Iterations 1 to 4


src/components/AddFoodForm.js
```diff
@@ -0,0 +1,51 @@
...
+            <Divider>Add Food Entry</Divider>
+
+            <label>Name</label>
+            <Input value={name} type="text" onChange={(e) => { setName(e.target.value) }} />
+
+            <label>Image</label>
+            <Input value={image} type="text" onChange={(e) => { setImage(e.target.value) }} />
+
+            <label>Calories</label>
+            <Input value={calories} type="text" onChange={(e) => { setCalories(e.target.value) }} />
```
TA-Remote, line 41: for Calories and Serving, you need to change the type to number. in this way you can prevent to getting other input(instead of number) from user and your code will not break ,

PR #1266, opened 2022-07-28
WDFTRT JUNE 2022 | Ruben Andino

rubenanlo:
ok, so.... it took me a while and now I start having the hang of it. I was sweating with the search one! I had to redo most of my code to make it work with the revised array. 

Felipe, sorry.... I didn't use the useEffect... 

Going to sleep now. Good night!! ;)

And, as always, thanks for reviewing this!!!!


rubenanlo:
Damn! just saw that the search feature is not updating when i delete a character.... Anyway, I hope this is the only bug you find :)

TA-Remote (commented):
Great Ruben! Left some notes to improve :)

PR #1268, opened 2022-07-28
Stephanie - Lab done

Stephanie-Tovar:


src/App.js
```diff
@@ -1,25 +1,37 @@
...
 import './App.css';
+import foods from './foods.json';
+import { useState } from "react";
+import Foodbox from './components/Foodbox';
+import AddFoodForm from './components/AddFoodForm';
+import Search from './components/Search';
 
 function App() {
+  const [foodsArr, setFoodsArr] = useState(foods);
+  const copy = [...foods];
```
TA-Remote, line 10: If you noticed, your search functionality works, but it does not retun the original array, the newly created items disappear.

- In you app.js you have only one state -> `foodsArr`
- At the moment, when you filter you use the `copy` variable to filter foods
- And when you delete letter from search input field you are still filtering that same foods array ( `copy` ), which is copy of the original array from db
- The idea here is to have two separate states. One to hold the full original list of foods. And the second one to hold foods to display. When you filter -> you Always filter the original array. And to dispaly -> you Always display that filtered array

TA-Remote (commented):
Great lab Stephanie! Left some notes to improve :)

PR #1272, opened 2022-07-28
Tobias Conrad WDFTJUN22

maktooob:
lab done, no bonus. need way more practice on this stuff :)


src/App.js
```diff
@@ -1,23 +1,47 @@
...
+    const newList = prevFoods.filter((element) => {
+      return element.name !== name;
+    })
+    return newList;
+  })
+}
+
+function searchFilter(input){
+  setFoods((prevFoods) => prevFoods.filter((e) =>  e.name.toLowerCase().includes(input)))
+}
```
TA-Remote, line 27: If you noticed, your search functionality works, but it does not retun the original array, if there is no input!

- In you app.js you have only one state -> `foods`.
- At the moment, when you filter you use this array to do it, and update that same `foods` array
- And when you delete letter from search input field you are still filtering that same `foods` array, which might have only few or none items. 
- The idea here is to have two separate states. One to hold the full original list of foods. And the second one to hold foods to display. When you filter -> you Always filter the original array. And to dispaly -> you always display that filtered array

TA-Remote (commented):
Great job Tobias! 

It's good practice ot keep your files organised. To keep all the components you create  in the `components` folder :)

PR #1286, opened 2022-08-17
[FTLIS2022] Raquel

raqpoletto:


webLisbonTA (commented):
Great job on the lab! Just one small thing on the delete, you were deleting from the original array and not the display one

raqpoletto:
> Great job on the lab! Just one small thing on the delete, you were deleting from the original array and not the display one

Thanks, Xico. I will fix the issue.

PR #1287, opened 2022-08-17
is not complete

AdrianaLeitao:


src/components/Foods/Foods.jsx
```diff
@@ -0,0 +1,52 @@
...
+        <Searchbar searchResults = {searchResults} />
+        <AddFoodForm createFood = {createFood} />
+        <FoodBox key={food} foods={foods} clickToDelete = {deleteFood} />;
+
+        {displayFood.map((food, index) => {
+        return(
+        <div key={index}>
+            {/* <p>{food.name}</p>
+            <img src={food.image} alt="" width={120}/> */}
+            <FoodBox food = {food} />
```
webLisbonTA, line 43: Be careful while declaring your props name. Your code was not working because they had the same name

src/components/Foods/Foods.jsx
```diff
@@ -0,0 +1,52 @@
...
+      const updatedFoods = [food, ...foods];
+      setFood(updatedFoods);
+      setDisplayFood(updatedFoods);
+    }
+
+    const deleteFood = (foodName) => {
+      let filterFoods = foods.filter((food) => food.name !== foodName);
+      setFood(filterFoods);
+      setDisplayFood(filterFoods);
+    }
```
webLisbonTA, line 30: This delete function should be passed inside of the FoodBox component inside of the map

webLisbonTA (commented):
Good job on the lab! You were just missing the delete and the search-bar was almost done

PR #1288, opened 2022-08-17
[FTLIS072022] - Miguel - lab and bonus

tomarnogueiracoding:


webLisbonTA (commented):
Great job on the lab including the bonus!

PR #1293, opened 2022-08-18
[WDFT072022] John Imeson

johnimeson0:
search console logs fine, but does not change the list. delete  I am having some trouble with, I think I'm just missing something small or not seeing a syntax error


src/components/FoodBox/FoodBox.jsx
```diff
@@ -0,0 +1,26 @@
...
+            title={food.name}
+            style={{ width: 230, height: 300, margin: 10 }}
+          >
+            <img src={food.image} height={60} alt="food" />
+            <p>Calories: {food.calories}</p>
+            <p>Servings: {food.servings}</p>
+            <p>
+              <b>Total Calories: {food.calories * food.servings} </b> kcal
+            </p>
+            <Button onClick={deleteFood} type="primary"> Delete </Button>
```
webLisbonTA, line 20: Here, you should pass an anonymous function and inside call the deleteFood function with the food.name as an argument:
```
() => deleteFood(food.name)
```

src/App.js
```diff
@@ -1,24 +1,50 @@
...
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
+      <AddFoodForm createFood={createFood} />
+      <h3>Food List</h3>
+      <hr />
+      <Search searchResults={searchResults}/>
+    {foods.map((el) => {
```
webLisbonTA, line 42: Here, the filter function was working but you forgot to map over the displayFoods array

src/App.js
```diff
@@ -1,24 +1,50 @@
...
+
+
+  const createFood = (food) => {
+    const updatedFoods = [food, ...foods];
+    setFoods(updatedFoods);
+    setDisplayFoods(updatedFoods);
+  }
+
+  const deleteFood = (foodId) => {
+    let searchedFoods = foods.filter((food) => food._id !== foodId);
```
webLisbonTA, line 31: Also, here you should compare the food name since the food does not have an id:
```
food.name !== food
```

webLisbonTA (commented):
Good job on the lab. You just had some minor issues on the last two iterations but the logic was there

PR #1363, opened 2022-11-03
[PTBER072022] berndHavekost

hvkst:


src/App.js
```diff
@@ -1,25 +1,42 @@
-import logo from './logo.svg';
-import './App.css';
+import rawFoods from './foods.json';
+import { useState } from 'react';
+import { v4 as uuidv4 } from 'uuid';
+
+import Heading from './components/Heading';
+import Content from './components/Content';
+
+const foodsWithIds = rawFoods.map((food) => ({ ...food, _id: uuidv4() }));
```
trebeljahr, line 8: this is named _id whereas further below you name it id...

src/App.js
```diff
@@ -1,25 +1,42 @@
...
+    setSearchInput(event.target.value);
+  };
+
+  const addNewFood = (newFood) => {
+    const updatedFood = [...food, newFood];
+    setFood(updatedFood);
+  };
+
+  const handleClickDelete = (id) => {
+    const foodToDelete = food.findIndex((obj) => obj.id === id);
```
trebeljahr, line 24: probably this should be obj._id

src/App.js
```diff
@@ -1,25 +1,42 @@
...
+    const foodMinusOne = [...food];
+    foodMinusOne.splice(foodToDelete, 1);
```
trebeljahr, line 26: this is a weird way of setting state. It's a convoluted way to write filter.

trebeljahr:
```js
setFood(old => old.filter(({_id}) => _id !== id))
```

src/components/AddFoodForm.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const handleNameInput = (e) => {
+    setName(e.target.value);
+  };
+
+  const handleImageInput = (e) => {
+    setImage(e.target.value);
+  };
+
+  // ok, this got me confused... somehow the InputNumber-Component from AntDesign
```
trebeljahr, line 20: this is something that can be expected though. 

What I am trying to say is, that it's a component, who knows what's going on inside it, just because it has the same name as the input element from HTML doesn't mean that it's going to work the same way as onChange. If in doubt, look at the [docs](https://ant.design/components/input-number/#:~:text=%2D-,onChange,function(value%3A%20number%20%7C%20string%20%7C%20null),-%2D).

src/components/Content.jsx
```diff
@@ -0,0 +1,43 @@
...
+      <Divider style={{ fontSize: '32px' }} orientation="center">
+        Food List
+      </Divider>
+      {food.length === 0 && <FeedbackMessage {...{ getFoodBack }} />}
+      <Row justify="space-evenly">
+        {food
+          // eslint-disable-next-line array-callback-return
+          .filter((food) => {
+            if (searchInput === '') {
+              return food;
```
trebeljahr, line 18: filter functions should return a boolean value, not the food item itself. i.e. this only works because food is a truthy value, but wouldn't work if it ever were a falsy value.

trebeljahr: instead this whole statement could be done in a single boolean. 
```js
return searchInput === "" || food.name.toLowerCase().includes(searchInput.trim().toLowerCase())
```  
which could be even shortened to only the second half, because an empty searchInput "" is included in any food.name. i.e. every string includes the empty string.

trebeljahr: so for filtering you could just do this: 
```js
.filter((food) => food.name.toLowerCase().includes(searchInput.trim().toLowerCase()))
```

src/components/Content.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { Divider, Row } from 'antd';
+
+function Content({ food, searchInput, handleClickDelete, getFoodBack }) {
+  return (
+    <>
+      <Divider />
+      <Divider style={{ fontSize: '32px' }} orientation="center">
+        Food List
+      </Divider>
+      {food.length === 0 && <FeedbackMessage {...{ getFoodBack }} />}
```
trebeljahr, line 12: that pun :see_no_evil:

trebeljahr (commented):
good work, though it seems like you should check out how filter works some more. :+1:

src/components/Content.jsx
hvkst: Thank you for the Feedback!!! :)

src/components/Content.jsx
hvkst: Awesome!
I didn´t know how to do it and copy/pasted that from some old medium post...

PR #1364, opened 2022-11-04
WDPT Ber Patrick

Patrick-Groot:


src/App.js
```diff
@@ -1,23 +1,52 @@
...
+  
+
+  //usually this works, but at times you have to click multiple times for the item to be deleted - in those cases it goes one parentNode too high -> the first console.log gives
+  //the UL element instead, so no food gets found there -> Index -1.
+  //But when clicking on the same button again one or two times it works, so not copmpletely sure why this error occurs in rare occasions. I guess having ".parentNode" 5times 
+  //isn't the most elegant way, perhaps it's due to that
+
+  function handleClickDelete (event) {
+    const nameOfElemToDelete =
+      event.target.parentNode.parentNode.parentNode.parentNode.parentNode.id;
```
trebeljahr, line 25: pass the id into the handleClickDelete instead of the event! That way things become much more elegant. Because you can use the id to do the filtering on the foods!

src/components/AddFoodForm.js
```diff
@@ -0,0 +1,85 @@
...
+        <Input
+          name="servings"
+          value={state.servings}
+          type="number"
+          onChange={handleChange}
+        />
+
+        <button type="submit">Create</button>
+      </form>
+      {hideState && <button onClick={handleClickHide}>Add New Food</button>}
```
trebeljahr, line 81: this should be a ternary.

trebeljahr: hideState ? oneButton : otherButton

trebeljahr (commented):
good work! but try to refactor the parent.parent.parent logic to use an id passed from the listItem to the handleDelete function!

Patrick-Groot:
> good work! but try to refactor the parent.parent.parent logic to use an id passed from the listItem to the handleDelete function!

Thanks for the feedback! I've refactored it a bit

PR #1366, opened 2022-11-05
complete, no bonus yet

jdlms:


src/App.js
```diff
@@ -1,23 +1,46 @@
...
+
+  // const handleDelete = (e) => {
+  //   setFoods((old) => {
+  //     return old.filter((food) => food.key !== e.currentTarget.id);
+  //   });
+  // };
+
+  const handleDelete = (e) => {
+    let newArr = [...foodsList].filter(
+      (food) => food.key !== e.currentTarget.id
```
trebeljahr, line 31: just know, there is a slight difference between currentTarget and target. they are not always, exactly the same: https://stackoverflow.com/questions/10086427/what-is-the-exact-difference-between-currenttarget-property-and-target-property

src/components/FoodBox.jsx
```diff
@@ -0,0 +1,33 @@
...
+export const FoodBox = ({ foodsList, searchString, handleDelete }) => {
+  let filteredFoods = foodsList.filter((food) =>
+    food.name.toLowerCase().includes(searchString.toLowerCase())
+  );
+
+  return (
+    <>
+      {filteredFoods.map((food) => {
+        return (
+          <Col>
```
trebeljahr, line 12: this misses the key ;)

trebeljahr (commented):
good work :)

PR #1369, opened 2022-11-09
[WD-PT-12JUL22] - João Serafim

JoaoSerafim3001:


src/components/AddFoodForm.jsx
```diff
@@ -0,0 +1,76 @@
+import { useState } from 'react';
+import { Divider, Input } from 'antd';
+
+const defaultFood = {
+  name: '',
+  image: '',
+  calories: 0,
```
trebeljahr, line 7: just be aware that these don't stay numbers, they become strings after your input is done. Luckily in the way that the data is used, it doesn't matter, but it might break if this is extended on.

src/components/FoodBox.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const FoodBox = ({ food }) => {
+  return (
+    <Col>
+      <Card title={food.name} style={{ width: 230, height: 300, margin: 10 }}>
+        <img src={food.image} height={60} alt="food" />
+        <p>Calories: {food.calories}</p>
+        <p>Servings: {food.servings}</p>
+        <p>
+          <b>Total Calories: {food.calories * food.servings} </b> kcal
```
trebeljahr, line 11: specifically here, for a newly created food this does assume this is a number. Now you are lucky, since JS handles this somewhat gracefully for you. 
i.e. 
"10" * 5 would return 50. 
5 * "10" would return 50. 

However if you do, 
"10" + 1 
you'd be in trouble. 
image

trebeljahr (commented):
nice work, just be careful with numbers vs. strings from inputs ^^

PR #1377, opened 2022-11-16
done till 6

MarioBaltazar19:


src/components/AddFoodForm.jsx
```diff
@@ -0,0 +1,69 @@
...
+function AddFood(props) {
+  const {createFood} = props
+  const [name, setName] = useState('');
+  const [image, setImage] = useState('');
+  const [calories, setCalories] = useState(0);
+  const [servings, setServings] = useState(0);
+
+  const handleName = (e) => setName(e.target.value);
+  const handleImage = (e) => setImage(e.target.value);
+  const handleCalories = (e) => setCalories(e.target.checked);
```
webLisbonTA, line 14: When you create a new food the calories are not being updated because here you have `e.target.checked` instead of `e.target.value` change the checked to value and it will work!

src/App.js
```diff
@@ -1,25 +1,49 @@
...
+import React, { useState } from 'react';
+import foods from './foods.json';
+import FoodBox from './components/FoodBox';
+import AddFood from './components/AddFoodForm';
+import SearchBar from './components/SearchBar';
+import { Row, Col } from 'antd'
 
 function App() {
+  const [allFoods, setAllFoods] = useState(foods)
+
```
webLisbonTA, line 11: For the Search function to work properly you need to add a new `useState` like this:
`  const [showFoods, setShowFoods] = useState(foods);`

src/App.js
```diff
@@ -1,25 +1,49 @@
...
+import AddFood from './components/AddFoodForm';
+import SearchBar from './components/SearchBar';
+import { Row, Col } from 'antd'
 
 function App() {
+  const [allFoods, setAllFoods] = useState(foods)
+  
+  const createFood = (newFood) => {
+    const updatedFood = [newFood, ...allFoods]
+    setAllFoods(updatedFood)
```
webLisbonTA, line 14: Every time you create a food you need to update the `allFoods` like you are doing here.
But now since you have another `useState` that will be the result of the search you need to update it as well. 
So under the `setAllFoods(updatedFood)` add this line `setShowFoods(updatedFood)`

src/App.js
```diff
@@ -1,25 +1,49 @@
...
+  }
+
+  const searchFood = (filterfoods) => {
+    const searchedFood = allFoods.filter(food => 
+    food.name.toLowerCase().includes(filterfoods.toLowerCase()))
+    setAllFoods(searchedFood)
```
webLisbonTA, line 20: The `allFoods` will be the original array that will always contain every food. So here you want to update the `setShowFoods` instead of the `setAllFoods` like this:
`setShowFoods(searchedFood)`

src/App.js
```diff
@@ -1,25 +1,49 @@
...
-      </header>
+      <div className='headT'>
+        <AddFood createFood={createFood} />
+      </div>
+      
+        <SearchBar filterfoods={searchFood} />
+      
+        
+      <Row>
+        {allFoods.map((food) => {
```
webLisbonTA, line 37: Last but not least, for your search bar to work, you now need to map over the `showFoods` instead of the `dishes`:
`{showFoods.map((food) => {`

PR #1380, opened 2022-11-16
[LISFT202210] Nathalie Cazemajou

natcaze:


src/components/AddFoodForm.jsx
```diff
@@ -0,0 +1,53 @@
+import { Divider, Input } from 'antd';
+import { useState } from 'react';
+import React from 'react';
+
+function AddFoodForm(newFood) {
```
webLisbonTA, line 5: Your create form is not working because of this:
`function AddFoodForm( newFood ) {`
If you want to extract the `newFood` directly from the function you need to wrap it in `{}` like this:
`function AddFoodForm( { newFood } ) {`
If you do this small correction your create from will work!

webLisbonTA:
Hi Nathalie, try to do the rest of the iterations so you can practice a bit more, overall good job!

PR #1381, opened 2022-11-16
[LISFT202210] Raiza Garcia

raizagarcia:


src/App.js
```diff
@@ -1,23 +1,44 @@
...
+import Search from './components/Search';
+
 
 function App() {
+  const [dishes, setDishes] = useState(foods);
+  const [search, setSearch] = useState(foods);
+
+  const createFood = (newDishes) => {
+    const updatedDishes = [newDishes, ...dishes]
+    setDishes(updatedDishes)
```
webLisbonTA, line 15: Since now you have a new 'useState' that is the result of the 'Search', every time you create a new food you need to update both, so under the 'setDishes' add a line updating the 'setSearch' as well: 
` setSearch(updatedDishes)`

src/components/AddFoodForm.js
```diff
@@ -0,0 +1,59 @@
...
+    setServings(1);
+  };
+
+  return (
+    <div>
+        
+        <h3>Add New Food</h3>
+      <form onSubmit={handleSubmit}>
+        <label htmlFor="name">Name</label>
+        <Input name='name' type="text" onChange={{handleName}}/>
```
webLisbonTA, line 40: In every 'Input' you are calling the 'handleFunction' like this:
`<Input name='name' type="text" onChange={{handleName}}/>`
In this way you are wrapping the function in an object, and doing so the 'onChange' will not be able to run the function, this is why every time you create a food it appears without the name neither the image that you passed. To correct this you just need to take out one pair of '{' like this:
`<Input name='name' type="text" onChange={handleName}/>`

PR #1382, opened 2022-11-16
[LISFT202210] Nathalia Alves

esthevlana:
lab solved


src/App.js
```diff
@@ -1,25 +1,52 @@
...
 import './App.css';
+import React, {useState} from 'react';
+import foods from './foods.json';
+import FoodBox from './components/FoodBox'
+import AddFoodForm from './components/AddFoodForm'
+import Search from './components/Search'
 
 function App() {
+  const [dishes, setDishes] = useState(foods)
+
```
webLisbonTA, line 10: For the Search function to work properly you need to add a new useState:
`  const [showFoods, setShowFoods] = useState(foods);`

src/App.js
```diff
@@ -1,25 +1,52 @@
...
+
+
+    const createFood = (newDishes) => {
+      const updatedDishes = [newDishes, ...dishes]
+      setDishes(updatedDishes)
```
webLisbonTA, line 15: Every time you create a food you need to update the dishes like you are doing here.
But now since you have another useState that will be the result of the search you need to update it as well. 
So under the ` setDishes(updatedDishes)` add this line `setShowFoods(updatedDishes)`

src/App.js
```diff
@@ -1,25 +1,52 @@
...
+    }
+    
+    const searchFood = (searchQuery) => {
+    const filteredFood = dishes.filter((food) =>
+      food.name.toLowerCase().includes(searchQuery.toLowerCase())
+      );
+    
+      setDishes(filteredFood);
```
webLisbonTA, line 23: The `setDishes` will be the original array that will always contain every food. So here you want to update the `setShowFoods` instead of the `setDishes` like this:
`setShowFoods(filteredFood)`

src/App.js
```diff
@@ -1,25 +1,52 @@
...
+    };
+
+    const deleteFood = (name) => {
+      const filteredFood = dishes.filter((dishes) => dishes.name !== name);
+      setDishes(filteredFood);
```
webLisbonTA, line 28: Here, since you now have a new `useState` that is the result of the search, you will need to update it as well every time you delete a food. So under the ` setDishes(filteredFood);` add a line like this `setShowFoods(filteredFood)`

src/App.js
```diff
@@ -1,25 +1,52 @@
...
-        >
-          Learn React
-        </a>
-      </header>
+    <div className="container">
+      <Search filterFood={searchFood}/>
+      <AddFoodForm createFood={createFood}/>
+      <h1 className="h1container">Food List</h1>
+    <div className="App"> {
+      dishes.map((dish, index) => {
```
webLisbonTA, line 37: Last but not least, for your search bar to work, you now need to map over the `showFoods` instead of the `dishes`:
` showFoods.map((dish, index) => {`

webLisbonTA:
Great job on the lab Nathália!

PR #1383, opened 2022-11-16
[LISFT202210]MiguelQuiaios

MiguelQuiaios:


src/App.js
```diff
@@ -1,23 +1,46 @@
...
 import './App.css';
+import React, { useState } from 'react';
+import foods from './foods.json';
+import FoodBox from './components/FoodBox';
+import AddFoodForm from './components/AddFoodForm';
+import Search from './components/Search';
+
 
 function App() {
+  const [dishes, setDishes] = useState(foods)
```
webLisbonTA, line 10: For the Search function to work properly you need to add a new `useState`, so under the `useState` that you already have you can add this line:
` const [showFoods, setShowFoods] = useState(foods);`

src/App.js
```diff
@@ -1,23 +1,46 @@
...
+  const createFood = (newDishes)=>{
+    const edit = [newDishes, ...dishes]
+    setDishes(edit)
```
webLisbonTA, line 13: Every time you create a food you need to update the `dishes` like you are doing here.
But now since you have another `useState` that will be the result of the search you need to update it as well. 
So under the ` setDishes(edit)` add this line `setShowFoods(edit)`

src/App.js
```diff
@@ -1,23 +1,46 @@
...
+  }
+const searchFood = (searchQuery)=>{
+  const filteredFood = dishes.filter((food) => 
+  food.name.toLowerCase().includes(searchQuery.toLowerCase())
+  );
+  setDishes(filteredFood)
```
webLisbonTA, line 19: The `dishes` will be the original array that will always contain every food. So here you want to update the `setShowFoods` instead of the `setDishes` like this:
`setShowFoods(filteredFood)`

src/App.js
```diff
@@ -1,23 +1,46 @@
...
+}
+const deleteFood = (name)=>{
+  const filteredFood = dishes.filter((dishes) => dishes.name !== name);
+  setDishes(filteredFood)
```
webLisbonTA, line 23: Here, since you now have a new `useState` that is the result of the search, you will need to update it as well every time you delete a food. So under the ` setDishes(filteredFood);` add a line like this `setShowFoods(filteredFood)`

src/App.js
```diff
@@ -1,23 +1,46 @@
...
-        >
-          Learn React
-        </a>
-      </header>
+    <div className="container">
+        <Search filterFood={searchFood}/>
+        <AddFoodForm createFood= {createFood}/>
+      <div className="App">
+
+       { dishes.map((dish) => {
```
webLisbonTA, line 34: Last but not least, for your search bar to work, you now need to map over the `showFoods` instead of the `dishes`:
`{ showFoods.map((dish) => {`

webLisbonTA:
Great job on the lab Miguel!

PR #1384, opened 2022-11-16
[LISFT202210]Camila

camiSaraiva:


src/App.js
```diff
@@ -1,23 +1,45 @@
...
+import Search from './components/Search';
+
 
 function App() {
+  const [dishes, setDishes] = useState(foods)
+  const [showDishes, setShowDishes] = useState(foods);
+
+  const createFood = (newFood) => {
+    const updatedFood = [newFood, ...dishes]
+    setDishes(updatedFood)
```
webLisbonTA, line 15: Since now you have a new 'useState' that is the result of the 'Search', every time you create a new food you need to update both, so under the 'setDishes' add a line updating the 'setShowDishes' as well: 
` setShowDishes(updatedFood)`

src/App.js
```diff
@@ -1,23 +1,45 @@
...
+  }
+
+  const searchFood = (searchQuery) => {
+    const filteredFood = dishes.filter((food) => 
+    food.name.toLowerCase().includes(searchQuery.toLowerCase())
+    );
+    setDishes(filteredFood);
```
webLisbonTA: Here you want to update only the 'showDishes' not the 'dishes' so instead of:
`setDishes(filteredFood);`
Do:
`setShowDishes(filteredFood);`

src/App.js
```diff
@@ -1,23 +1,45 @@
...
-        >
-          Learn React
-        </a>
-      </header>
+    <div className='container'>
+      <Search filterFood={searchFood}/>
+      <AddFoodForm createFood = {createFood} className='crtForm'/>
+      <h1 className='tlt'>Food List</h1>
+      <div className="App">
+        {dishes.map((dish) => {
```
webLisbonTA: Now that you have a new 'useState' the is the result of the 'Search' you need to loop over it, so instead of:
` {dishes.map((dish) => {`
Do:
`  {showDishes.map((dish) => {`

webLisbonTA:
Great job on the lab Camila!

PR #1443, opened 2023-01-20
Update antd setup instructions

ross-u:
- Update setup instructions for `antd` v5. It is no longer required to import the antd stylesheet ([docs](https://github.com/ant-design/ant-design#-usage)).
- Minor formatting and style changes.


ross-u:
I updated the lab solution to reflect the change: https://github.com/ironhack-edu/webdev-labs-solutions-module-3/pull/1

PR #1477, opened 2023-03-01
[LISFT0123] Alexandre Álvaro

AleAlvo:
initial commit


src/App.js
```diff
@@ -1,40 +1,84 @@
...
+            <FoodBox
+              span={6}
+              food={food}
+              deleteFood={deleteFood}
+              key={food.name}
+            />
+          );
+        })}
       </Row>
+
```
joaopdg, line 79: Iteration 8 - if you want to show something when there are no foods left, and foods is an array, you can check the length of the array:

```
function App() {

  return (
    <div className="App">
             (...)
      { !showFood.length && 
        <p>No results found!</p> }
    </div>
)}
```

joaopdg (commented):
Great job on the lab Alexandre!

PR #1480, opened 2023-03-01
[LIS0123] Wagner

waggnerl:


src/App.js
```diff
@@ -1,24 +1,67 @@
...
+import { useState } from 'react';
+import FoodBox from './components/FoodBox';
+import AddFoodForm from './components/AddFoodForm';
+import FilterFoodForm from './components/FilterFoodForm';
 
 function App() {
+  const [foodsDefault, setFoodsDefault] = useState(foods);
+  const [displayDiv, setDisplayDiv] = useState(false);
+  const [foodsDefaultToUpdate, setFoodsDefaultToUpdate] = useState(foods);
+  const updateFoodsDefault = (newFood) => {
```
lzaquine, line 12: This isn't working because the `updateFoodsDefault` function is updating the `foodsDefault` state instead of `foodsDefaultToUpdate`, which is the state used to display the list of foods after filtering. To fix this, you can modify the `updateFoodsDefault` function to update `foodsDefaultToUpdate` instead of `foodsDefault`:

```
const updateFoodsDefault = (newFood) => {
  let copyFoodsDefaultToUpdate = [...foodsDefaultToUpdate];
  copyFoodsDefaultToUpdate.push(newFood);
  setFoodsDefaultToUpdate(copyFoodsDefaultToUpdate);
};
```

lzaquine (commented):
Great job on the lab, Wagner!

PR #1482, opened 2023-03-01
[LISFT0123] Maria Carvalho

miacsilva:


src/App.js
```diff
@@ -1,23 +1,60 @@
...
+  const addToList = (food) => {
+    const foodsCopy = [food, ...foodsJSON];
+    setFoods(foodsCopy);
+  }
+
+  const searchFoods = (query) => {
+    const filteredFoods = showFoods.filter((food) =>
+      food.name.toLowerCase().includes(query.toLowerCase())
+    );
+    setShowFoods(filteredFoods);
```
lzaquine, line 30: Here, you should've called `setFoods` instead of `setShowFoods` because the initial state of `showFoods` is the same as `foods`, which is the state that is being modified when you filter the results.

```
const searchFoods = (query) => {
    const filteredFoods = showFoods.filter((food) =>
      food.name.toLowerCase().includes(query.toLowerCase())
    );
    setFoods(filteredFoods);
  };
```

src/App.js
```diff
@@ -1,23 +1,60 @@
...
+
+    const deleteFood = (foodName) => {
+    const filteredFoods = foods.filter((food) => {
+      return food.name !== foodName; 
+    });
+    setFoods(filteredFoods);
+  };
+  
+  const addToList = (food) => {
+    const foodsCopy = [food, ...foodsJSON];
```
lzaquine, line 22: You are creating a copy of the `foods` array. In the `addToList` function, you are currently using `foodsJSON` which is the original data and not the updated state `foods`. Instead, you should use `foods` to create a copy of the current state and add the new food item to it.

```
const addToList = (food) => {
    const foodsCopy = [food, ...foods];
    setFoods(foodsCopy);
  };
```

lzaquine (commented):
Great job on the lab, Maria!

PR #1483, opened 2023-03-01
[LISFT0123] Déborah

deborahdelimaa:


src/App.js
```diff
@@ -1,24 +1,63 @@
...
 
 function App() {
+  const foodArr = [...foodsJSON]
+  const [foodList, setFoodList] = useState(foodsJSON);
+  const [showFood, setShowFood] = useState(foodsJSON)
+  
+
+  const addFood = (food) => {
+    const foodListCopy = [...foodList, food];
+    setFoodList(foodListCopy)
```
joaopdg, line 19: Your `addFood()` function is not working because you're only updating `foodList` with `setFoodList` you need to update `showFood` as well, because that's the variable that you're using to loop through and display all the foods.
```
  const addFood = (food) => {
    const foodListCopy = [food, ...foodList];
    setFoodList(foodListCopy);
    setShowFood(foodListCopy);
  };
```

src/App.js
```diff
@@ -1,24 +1,63 @@
...
+  };
+
+  const searchFood = (query) => {
+    const filteredFood = foodsJSON.filter((food)=> 
+    foodsJSON.name.toLowerCase().includes(query.toLowerCase())
+    );
+    setShowFood(filteredFood)
```
joaopdg, line 26: Your `searchFood` is not working because:
- you're filtering through `foodsJSON` (json file) instead of `foodList` (variable);
- inside the filter you're checking the `.name` of `foodsJSON` (json file) instead of checking the name of every single element in the `foodList` array with `food` (placeholder);
- you're using `.includes()` an `array` method to compare two `strings`;
- you're only updating `showFood` with `setShowFood` you need to update `foodList` as well;

```
  const searchFood = (query) => {
    const filteredFood = foodList.filter((food) =>
    food.name.toLowerCase().startsWith(query.toLowerCase()));
    setShowFood(filteredFood);
    setFoodList(filteredFood);
  };
```

joaopdg (commented):
Great job on the lab Déborah!

src/App.js
```diff
@@ -1,24 +1,63 @@
...
+      return (
+        <div>
+
+
+  <p> {food.name}</p>
+  <img src={food.image} width={90} alt=""/>
+</div>
+      )
+    })
+    }
```
joaopdg, line 50: Delete this `.map()` you're not supposed to render the list of food two times, in the lab instructions they ask you to replace this loop with the one that you have below that is already styled.

src/App.js
```diff
@@ -1,24 +1,63 @@
...
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
+    {foodArr.map((food)=> {
+      return (
+        <div>
```
joaopdg, line 42: Delete this `.map()` you're not supposed to render the list of food two times, in the lab instructions they ask you to replace this loop with the one that you have below that is already styled.

PR #1485, opened 2023-03-01
[LISFT0123] André Lisboa

andrelisboapt:
lab with bonus done


src/App.js
```diff
@@ -1,25 +1,74 @@
...
+  const addToList = (item) => {
+    const foodCopy = [item, ...food];
+
+    setFood(foodCopy);
+  };
+
+  const searchFood = (query) => {
+    const filteredFood = food.filter((item) =>
+      item.name.toLowerCase().includes(query.toLowerCase())
+    );
```
joaopdg, line 24: Your search bar is not working because you're using `.includes()` that is an array method.

You want to compare `item.name` with `query` both are `strings` so `.includes()` will never work here unless you turn those `strings `into `arrays`.

Or you can simply use `.startsWith()` that is a method that compares two `strings`:
```
item.name.toLowerCase().startsWith(query.toLowerCase())
```

src/App.js
```diff
@@ -1,25 +1,74 @@
...
 function App() {
+  const [food, setFood] = useState(foods);
+  const [alwaysFood, setAlwaysFood] = useState(foods);
+  const [visible, setVisible] = useState(true);
+  const [isFoodArrayEmpty, setIsFoodArrayEmpty] = useState(false);
+
+  const addToList = (item) => {
+    const foodCopy = [item, ...food];
+
+    setFood(foodCopy);
```
joaopdg, line 18: Your create food is not working because here you're updating `food` with `setFood` but the variable that you're using to loop through and display all the foods is `alwaysFood` so you need to update the `alwaysFood` with `setAlwaysFood` as well:
```
  const addToList = (item) => {
    const foodCopy = [item, ...food];
    setFood(foodCopy);
    setAlwaysFood(foodCopy);
  };
```

src/App.js
```diff
@@ -1,25 +1,74 @@
...
-      </header>
+      <button onClick={toggleVisibility}>
+        {visible ? 'Hide Form' : 'Add New Food'}
+      </button>
+      {visible && <AddFoodForm addToList={addToList} />}
+      <Searchbar searchFood={searchFood} />
+
+      {isFoodArrayEmpty ? (
+        <p>oops ta fdd</p>
+      ) : (
```
joaopdg, line 53: This message is not being displayed when there are no foods left. A easier way of doing this is by checking the `length` of the foods array, if it is empty display the message:
```
{!alwaysFood.length && <p>oops ta fdd</p>}
```

joaopdg (commented):
Great job on the lab André!

PR #1486, opened 2023-03-01
[LISFT0123] Tiago Rato

TiagoRato13:


src/App.js
```diff
@@ -1,23 +1,54 @@
...
-      </header>
+      <AddFoodForm addToList={addToList} />
+      <Button> Hide Form / Add New Food </Button>
+
+      <Search searchRecipe={searchRecipe} />
+
+      <Divider>Food List</Divider>
+
+      <Row style={{ width: '100%', justifyContent: 'center' }}>
+        {showFood.map((food) => {
```
brunorocha20, line 48: You are updating your `setSearchFood` in the `deleteRecipe` function, but when doing the `.map` to show the food you have, you are using the `showFood` state variable that was not changed after you click the button to delete the food.

src/App.js
```diff
@@ -1,23 +1,54 @@
...
+    setShowFood(filteredRecipes);
+  };
+
+  //delete function
+
+  const deleteRecipe = (foodName) => {
+    const filteredRecipes = showFood.filter((recipe) => {
+      return recipe.name !== foodName;
+    });
+    setSearchFood(filteredRecipes);
```
brunorocha20, line 35: To prevent the error below from happening, you could just update both State "sets".
So, you could just add this:
`setShowFood(filteredRecipes);`

brunorocha20 (commented):
Good job on the lab Tiago.

PR #1490, opened 2023-03-01
[LISFT0123] João Carneiro

joao621119:


src/App.js
```diff
@@ -1,23 +1,70 @@
...
+  }
+  
+  const searchRecipes = (query) => {
+    const filteredRecipes = foodsJSON.filter((recipe) =>
+      recipe.name.toLowerCase().includes(query.toLowerCase())
+    );
+    setFoodList(filteredRecipes);
+  };
+
+  const deleteRecipe = (recipe) => {
```
lzaquine, line 26: Here, in the `deleteRecipe` function, you are using the `foodsJSON` array to filter out the recipe to delete but this is not updated when a recipe is added or deleted. Instead, you should use the `foodList` state variable that you are updating when a recipe is added or filtered. So you need to replace `foodsJSON` with `foodList` in the `deleteRecipe` function.

```
const deleteRecipe = (recipe) => {
  const filteredRecipes = foodList.filter((food) => {
    return food.name !== recipe;
  });
  setFoodList(filteredRecipes);
};
```

lzaquine (commented):
Great job on the lab, Joao!

PR #1492, opened 2023-03-01
[LISFT0123] Caio Mazur

caiomazur:


src/App.js
```diff
@@ -1,25 +1,59 @@
...
+      food.name.toLowerCase().includes(query.toLowerCase())
+    );
+    setShowFood(filteredFoods);
+  };
+
+  const deleteFood = (name) => {
+    const filteredFoods = foods.filter((food) => {
+      return food.name !== name;
+    });
+    setShowFood(filteredFoods);
```
joaopdg, line 30: Your delete food is not working properly because:
- you're filtering through `foods` (json file) instead of `food` (variable);
- you're only updating `showFood` with `setShowFood` you need to update `food` as well;
```
  const deleteFood = (name) => {
    const filteredFoods = food.filter((food) => {
      return food.name !== name });
    setShowFood(filteredFoods);
    setFood(filteredFoods);
  };
```

joaopdg (commented):
Great job on the lab Caio!

PR #1493, opened 2023-03-01
[LISFT0123] Erika

erikasoaress:
Not able to finish the lab. Felt frustrated because I understood the class but I don't feel able to apply the content to the labs many times...

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,64 @@
-import logo from './logo.svg';
+// src/App.js
 import './App.css';
+import foods from "./foods.json";
+import { useState } from 'react';
+import FoodBox from './components/FoodBox';
+import AddFoodForm from './components/AddFoodForm';
+import Searchbar from './components/Searchbar';
+
+
 
 function App() {
-  return (
-    <div className="App">
-      <header className="App-header">
-        <img src={logo} className="App-logo" alt="logo" />
-        <p>
-          Edit <code>src/App.js</code> and save to reload.
-        </p>
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+  const [foodList, setFoodList] = useState(foods);
+
+  const addFood = (food) => {
+    const foodListCopy = [...foodList, food]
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -1,25 +1,64 @@
...
+
+    <AddFoodForm addFood={addFood} 
+    />
+
+    {/* <Searchbar searchFoods={searchFoods} /> */}
+
+    <Searchbar searchFoods={searchFoods} />
+      {foods.map((food) => {
+        return <FoodBox food={food} deleteFoods={deleteFoods} key={food.name} />;
+      })}
```
joaopdg, line 47: Delete this block of code:
```
      {foods.map((food) => {
        return <FoodBox food={food} deleteFoods={deleteFoods} key={food.name} />;
      })}
```
You're not supposed to render the list of foods two times, besides you're using `foods`(json file) to loop through and you should use `foodList` the variable that you declared with the `useState`

src/App.js
```diff
@@ -1,25 +1,64 @@
...
+
+    {foodList.map((food) => {
+      return <FoodBox food={food} key={food.name}/>;
```
joaopdg, line 50: Pass the `deleteFoods()` function as `props` to the `FoodBox` component:
```
    {foodList.map((food) => {
      return <FoodBox food={food} key={food.name} deleteFoods={deleteFoods}/>;
    })}
```

src/App.js
```diff
@@ -1,25 +1,64 @@
...
+    food.name.toLowerCase().includes(query.toLowerCase())
+    );
+    setFoodList(filteredFoods);
+  };
+
+  const deleteFoods = (foods) => {
+    const deletedFoods = foods.filter((food) => {
+      return food.name !== foods;
+    });
+    setFoodList(deletedFoods);
```
joaopdg, line 30: You're `deleteFoods()` function is not working because you're filtering `foods`(json file) instead of filtering `foodList`(variable):
```
  const deleteFoods = (foods) => {
    const deletedFoods = foodList.filter((food) => {
      return food.name !== foods;
    });
    setFoodList(deletedFoods);
  };
```

joaopdg (commented):
Great job on the lab Erika, you only had a little mistake in the deleteFoods() function.

PR #1494, opened 2023-03-01
Eveline Coenegrachts

evelinecoen:


src/App.js
```diff
@@ -1,25 +1,63 @@
...
+        )})}
+
+
+
+      </div>
     </div>
   );
 }
-
+}
```
lzaquine, line 62: You have this extra `}` that should be closing the deleteFood instead on line 33.

src/App.js
```diff
@@ -1,25 +1,63 @@
...
+  food.name.toLowerCase().includes(query.toLowerCase())
+  );
+  setAllFoods(filteredFoods);
+};
+
+const deleteFood = (foodId) => {
+  const filteredFoods = allSearchFood.filter((manyFood) => {
+    return manyFood.name !== foodId;
+  });
+  setAllSearchFood(filteredFoods)
```
lzaquine, line 32: Here, the deleteFood function is not working when `setAllSearchFood` is called because it uses the `allSearchFood` state to filter out the food to be deleted. The `allSearchFood` state is only updated when the `searchFood` function is called, and it is not updated in the deleteFood function.

This should fix it:
```
const filteredFoods = allFoods.filter((manyFood) => {
  return manyFood.name !== foodId;
});
```

lzaquine (commented):
Great job on the lab, Eveline!

PR #1497, opened 2023-03-01
[LIS0123] Marisha Deroubaix

Deroubaix:


src/App.js
```diff
@@ -1,23 +1,66 @@
...
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
+      <div>
+
+        <AddFoodForm createFood={addNewFood}/>
```
brunorocha20, line 45: Here, you need to send the `addNewFood` as props to your component where you are getting the function to create the food and pass the correct values. So, you need to do it like this:
`<AddFoodForm addNewFood={addNewFood}/>`

brunorocha20 (commented):
Great job on the lab Marisha.

PR #1501, opened 2023-03-02
[LISFT0323] Assebe Kleinert

Assebe:


src/components/AddFoodForm.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const handleName = (e) => setName(e.target.value);
+  const handleImage = (e) => setImage(e.target.value);
+  const handleCalories = (e) => setCalories(e.target.value);
+  const handleServing = (e) => setServing(e.target.value);
+
+  const handleSubmit = (e) => {
+    e.preventDefault();
+
+    const newFood = { name, image, calories, serving };
```
joaopdg, line 20: When you create a new food `servings` is empty and `total calories` is `NaN` because you're using `serving` instead of `servings` you're missing a `s`

src/App.js
```diff
@@ -1,23 +1,49 @@
...
+    setFood(foodCopy);
+    setFindFood(foodCopy);
+  };
+
+
+  const deleteFood = (foodId) => {
+    const filteredFood = foods.filter((dish) => {
+      return foods._id !== foodId;
+    });
+    setFood(filteredFood);
```
joaopdg, line 31: Your delete food is not working because:
- you're filtering through `foods` (json file) instead of `food` (variable);
- there is no `_id` in food object to compare, you can use `name` instead;
- you're only updating `food` with `setFood` you need to update `findFood` as well;
- you're passing the `deleteFood()` function as `props` to `FoodListComplete` but inside this component you're not using this function, you're using a copy of the function that you pasted inside the component;

App.js:
```
  const deleteFood = (foodId) => {
    const filteredFood = food.filter((dish) => {
      return dish.name !== foodId.name });
    setFood(filteredFood);
    setFindFood(filteredFood);
  };
```

FoodListComplete Component:
```
function FoodListComplete(props) {
  const { deleteFood, findFood } = props;

  return (
    <div>
       ( ... )
      <button onClick={() => deleteFood(dish)}>Delete</button>
    </div>
  );
}
export default FoodListComplete;
```

joaopdg (commented):
Great job on the lab Assebe!

PR #1502, opened 2023-03-02
[LISFT0123] Rafaela Umbelino

RafaelaUmbelino:


src/App.js
```diff
@@ -1,25 +1,58 @@
...
+import AddFoodForm from './components/AddFoodForm';
 
 function App() {
+  
+  const [list, setList] = useState(foods);
+
+  const addFood = (food) => {
+    const foodListCopy = [...list, food]
+    setList(foodListCopy)
+
```
brunorocha20, line 14: You need to close this `addFood` function before this return below. Otherwise, it is not reading the values in your `return` correctly.

src/App.js
```diff
@@ -1,25 +1,58 @@
...
+  {/* //       return (
+  //         <div key={el.name} >
+  //           <p>{el.name}</p>
+  //           <img src={el.image} alt="foodImage" className="src" width={100} />
+  //         </div>
+  //       );
+  //     })} */}
+
+  <AddFoodForm addFood={addFood} />
+    {food.map((food) => {
```
brunorocha20, line 27: Here, you need to map over the State variable that stores your data.
So, in this case, you needed to map over the `list` variable that you defined earlier.

brunorocha20 (commented):
Good job Rafaela, check the annotations for a better understanding.

PR #1589, opened 2023-05-24
Nuno Durão [WD-FT-LIS-APR-2023]

NunoDurao:


src/App.js
```diff
@@ -1,25 +1,50 @@
...
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
+      <h1>Food List</h1>
+      <Search handleSearch={handleSearch} />
```
joaopdg, line 35: Here you're passing props with the name `handleSearch` but inside the component you calling it `onSearch` change the name of the props here and your search bar will work:
```
<Search onSearch={handleSearch} />
```

src/App.js
```diff
@@ -1,25 +1,50 @@
...
+      <div className="container">
+        <div className="food-list">
+          {filteredFoodList.map((food, index) => (
+            <FoodBox key={index} food={food} onDelete={() => handleDelete(index)} />
+          ))}
+        </div>
+        <AddFoodForm handleAddFood={handleAddFood} />
```
joaopdg, line 42: Same thing as above, you're passing props with the name `handleAddFood` but inside the component you calling it `onAddFood` change the name of the props here and your create food will work:
```
<AddFoodForm onAddFood={handleAddFood} />
```

PR #1594, opened 2023-05-24
LIS-FT-APR23 Guy-Dagan

Guy-Dagan:


src/App.js
```diff
@@ -1,25 +1,42 @@
...
+import AddFoodForm from './Components/AddFoodForm';
+import Search from './Components/Search';
 
 function App() {
+  const [foods, setFoods] = useState(foodsData);
+  const [filteredFoods, setFilteredFoods] = useState(foodsData);
+
+  const addNewFood = (newFood) => {
+    const updatedFoods = [...foods, newFood];
+    setFoods(updatedFoods);
```
brunorocha20, line 16: When you create a new food on the website nothing appears, and the function to `addNewFood` is correct. The only thing that you are missing is that you also need to update the `setFilteredFoods()` with the `updatedFoods` that you created, because that's the one that you are using the `.map` method to render the list of all the foods. 
The way you did, you are only updating the `foods` array, and not the filtered array that you are showing to the user.

brunorocha20 (commented):
Good job on the lab.

PR #1597, opened 2023-05-25
REVIEW [WDFT LIS APR23] Victoria Avelar

avelarvicky:


src/Components/AddFoodForm.jsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <div>
+      <form onSubmit={handleSubmit}>
+        <Divider>Add Food Entry</Divider>
+
+        <label>Name</label>
+        <Input value={name} type="text" name="name" onChange={handleNameInput} />
+
+        <label>Image</label>
+        <Input value={image} type="text" name="name" onChange={handleNameInput} />
```
joaopdg, line 41: Here you're using `handleNameInput` instead of using `handleImageInput` thats why the image is not working

src/App.js
```diff
@@ -1,23 +1,57 @@
...
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
+    <div>
+      {foodList.map((food) => (
```
joaopdg, line 37: You're rendering the foodlist 2 times, you only need to render it 1 time with the FoodBox component

PR #1599, opened 2023-05-25
LIS-WD-FT-APR-23-TOMÁS-MARIA

SinTomas:


src/Components/Search.jsx
```diff
@@ -0,0 +1,23 @@
+import {Divider, Input} from 'antd'
+import {useState} from 'react';
+
+function Search(props) {
+  let [searchFood, setSearchFood] = useState("");
+
+  let handleSelect = (e) => {
+    let inputSearch = e.target.value.toLowerCase()
+    setSearchFood(inputSearch);
+    props.search(inputSearch);
```
joaopdg, line 10: the props that you're passing to this component is called `searchFood` and not `search`

src/App.js
```diff
@@ -1,25 +1,50 @@
...
+        return food.name.toLowerCase().includes(str.toLowerCase());
+      });
+    }
+    setFoodFilterData(filteredFood);
+  };
+
+  return(
+      <div className="App">
+        <Search searchFood={filterFoodList}/>
+        {foods.map((food) => {
```
joaopdg, line 36: Now that you have the search bar done you need to map `foodFilterData` instead of `foods`

PR #1607, opened 2023-05-25
lab done

ana-rak:
- search bar and the delete button are not functional

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,23 +1,33 @@
-import logo from './logo.svg';
 import './App.css';
+import { useState } from 'react';
+import foods from './foods.json';
+import FoodBox from './FoodBox/index';
+import AddFoodForm from './AddFoodForm';
+import Search from './Search';
+
 
 function App() {
+  const [foodList, setFoodList] = useState(foods);
+  const addFood = (food)=>{
+    const updateFoods = [food, ...foodList];
+    setFoodList(updateFoods);
+  };
+  const handleSearch = (query) => {
+    const filtered = foodList.filter((food) =>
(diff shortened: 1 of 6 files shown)


src/FoodBox/index.jsx
```diff
@@ -0,0 +1,23 @@
...
+    <div className="card">
+      <Col>
+        <Card title={food.name} style={{ width: 230, height: 300, margin: 10 }}>
+          <img src={food.image} height={60} alt="food" />
+          <p>Calories: {food.calories}</p>
+          <p>Servings: {food.servings}</p>
+          <p>
+            <b>Total Calories: {food.calories * food.servings} </b> kcal
+          </p>
+          <Button type="primary"> Delete </Button>
```
brunorocha20, line 16: You have a delete button in your FoodBox component, however, you are not handling that event at all, so, your delete is not functional.

src/App.js
```diff
@@ -1,23 +1,33 @@
...
 
 function App() {
+  const [foodList, setFoodList] = useState(foods);
+  const addFood = (food)=>{
+    const updateFoods = [food, ...foodList];
+    setFoodList(updateFoods);
+  };
+  const handleSearch = (query) => {
+    const filtered = foodList.filter((food) =>
+      food.name.toLowerCase().includes(query.toLowerCase())
```
brunorocha20, line 17: Even if this component was working, after you filter the food, you are not updating the food list that the user sees, so, it is like the component doesn't work.

brunorocha20 (commented):
Good job.

PR #1696, opened 2023-09-08
[PT] Marcos & Franklin

franklinosei:


src/App.js
```diff
@@ -1,23 +1,66 @@
...
+  });
+
+  const [foodList, setFoodList] = useState(foods);
+  const [filteredFoodList, setFilteredFoodList] = useState(foodList);
+
+  const [showForm, setShowForm] = useState(false);
+
+  function addNewFood(newFood) {
+    const newFoodList = [...foodList, newFood];
+    setFoodList(newFoodList);
```
Web-Dev-PT-EN, line 25: Here you probably need to also add to the other state at the same time, so it would appear in the list

Web-Dev-PT-EN (commented):
Good work

PR #1699, opened 2023-09-09
done by Nermeen and Maria

nermeenmostafa:


src/App.js
```diff
@@ -1,7 +1,42 @@
...
+  }
+
+  const searchFood = (str) => {
+    let searchedFood
+
+    if(str === '' ){
+      searchedFood = foodMaster
+    } else {
+      searchedFood = foodMaster.filter((food) =>{
+        return food.name.toLowerCase() === str.toLowerCase()
```
iPoolito: It would be better to use the include method so actually would start looking for the string and not specting to be exacting the same 

```suggestion
        return food.name.toLowerCase().includes(str.toLowerCase())
```

src/App.js
```diff
@@ -1,7 +1,42 @@
...
+  function deleteFood(foodName){
+    const filteredFoods = foods.filter(
+      (oneFood) =>  oneFood.name !== foodName 
+    )
+    setAllFoods(filteredFoods)
+  }
+
   return (
     <div className="App">
       <header className="App-header">
```
iPoolito: I would remove the header because its only the template of react so we currently dont need it

src/App.js
```diff
@@ -18,6 +53,24 @@ function App() {
           Learn React
         </a>
       </header>
+    <SearchBar searchFood={searchFood}/>
+    <AddFoodForm addNewFood={addNewFood}/> 
+            <h1>Foodlist</h1>
+            <div className='container'>
+          {foods.map((oneFood) => {
+        return (
+          <div>
```
iPoolito: Remember when we use map its always need to put the key property to highest component 
```suggestion
          <div key={oneFood.name}> 
```

iPoolito:
Good job 🚀

PR #1715, opened 2023-09-17
Xavier

Xavieros03:


src/App.js
```diff
@@ -1,23 +1,41 @@
...
+    return food;
+  });
+
+  const [foodList, setFoodList] = useState(foods);
+  const [filteredFoodList, setFilteredFoodList] = useState(foodList);
+
+
+  function addNewFood(newFood) {
+    const newFoodList = [...foodList, newFood]
+    setFoodList(newFoodList)
```
iPoolito, line 22: here also i would set the filteredFoodList in order to have also displayed the new food that you add
```suggestion
    setFoodList(newFoodList)
        setFilteredFoodList(newFoodList)
```

src/App.js
```diff
@@ -1,23 +1,41 @@
...
-        >
-          Learn React
-        </a>
-      </header>
+      <AddFoodForm addNewFood={addNewFood} />
+
+      <h1>Food List</h1>
+      <Row>
+        {filteredFoodList.map((food) => {
+          return <FoodBox food={food} deleteHandler={deleteHandler}></FoodBox>;
```
iPoolito, line 36: Remember when we use map the highest component should have a key prop in order to react know which node its creating and have the correct reference
```suggestion
          return <FoodBox food={food} deleteHandler={deleteHandler} key={food.name}></FoodBox>;
```

PR #1728, opened 2023-11-02
[LIS-FT-OCT23] Fares Qaddoura

FaresQ29:
All iterations were green at first, but when I implemented the search bar it made the second iteration fail but everything still works as intended


jacobviv (commented):
Great work Fares,
tests are not happy but your app runs smooth!!

PR #1735, opened 2023-11-02
[LIS-FT-OCT23] João_Cravidão

joaocravidao:


webLisbonTA:
Hi João, 
Great job on your job!
The fact that we create a handleFunction for each element (name, image, calories and servings) helps us to set the values to an empty string, after we click on "Create". 

Overall, great job!

PR #1746, opened 2023-11-09
[FTWDOCT23]Hugo Pereira - Mahmoud

Mahmoud11111111:


hugocastroferro:
Knew until IT5, it 6 i didnt knew nothing at was, was really confused because im tired.
mahmoud did it6 alone because he knew what to do and i didnt.

PR #1753, opened 2023-11-09
[FWD-FT-October2023] Fran

Franmg94:
I was in pair with Huseyin but he couldn't do the lab in pairs today.

I didn't manage to finish the iteration 5, I got stack in passing the information from the inputs.


TA-Remote:
hey @Franmg94 
well done with your iterations🙌

To keep separated the 'concerns' of components, we create state variables where they get 'used' or consume the most. 
In the case of the Form, as the state variables for each input are needed just inside the Form component, we can create them there. Same with the handleSubmit function that will take care of creating the newFood obj from the state values.

The logic to update the foodList instead can stay in App.jsx ( as the foodList state is in App.jsx ).
As you did for the deleteFoodObj, you can create your addFoodObj function, that's expecting to receive the 'newFoodObj' as parameter. This function will update the state --> setFoodList(). And pass this function as callback via props to the Form component, where you can invoke it, inside handleSubmit, passing the just created newFood as argument.

A way that I use to try to understand where to create the state is: as a convention you hardly see passing the setState functions via props. When you need to do that, meand your state variables are probaly not in the rigth place, or at least not the most convenient😉

Keep it up with you gteat job!!!
Ale

Franmg94:
Thank you for your feedback Ale! 

As soon as I have energy and time I will try to apply this and improve it :)

